PR #1, opened 2018-07-22
Open source everything

ivanvanderbyl:
> WIP

This PR open sources and breaks out the packages which provide @flood/element and @flood/cli


ivanvanderbyl:
In our examples and documentation I think we should always give the most simple usage:

```typescript
import { suite } from '@flood/element'

export default suite(step => {
	step('Test: Start', async browser => {
		await browser.visit('https://challenge.flood.io/')
		await browser.takeScreenshot()
	})
})
```

ivanvanderbyl:
I wonder if it's possible to declare in typescript that `export const settings = {}` is of type `TestSettings`?

PR #15, opened 2018-10-03
WIP custom http headers

lachie:
Simple facility for setting custom http headers.
WIP because I'm not quite sure how to test.

Need a repeatable site like `challenge.flood.io` to test this from a smoke test.
I tried setting User-Agent, but you need to do that via `setUserAgent`.

### DIFF
--- a/packages/element/src/runtime/Browser.ts
+++ b/packages/element/src/runtime/Browser.ts
@@ -503,6 +503,11 @@ export class Browser<T> implements BrowserInterface {
 		return this.page.setUserAgent(userAgent)
 	}
 
+	@rewriteError()
+	public async setExtraHTTPHeaders(headers: { [key: string]: string }): Promise<void> {
+		return this.page.setExtraHTTPHeaders(headers)
+	}
+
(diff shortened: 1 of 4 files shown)


ivanvanderbyl:
We could test with setting content-type on something and making it return JSON?

ivanvanderbyl:
I think we should ship this as is, WDYT @lachie ?

mattisa:
Will this be merged?
Just waiting this change to bypass some stuff on load testing.

90kts:
@lachie http://flooded.io/headers will echo back any request headers

90kts:
Note you'll also need to bust past the varnish cache on that endpoint for testing with unique parameter e.g. `curl -H "Special: Sauce" "http://flooded.io/headers?param=$(date +%s%N)"`

jiffypopjr:
Would love to use this, what is holding it up from the merge?!

ivanvanderbyl:
@jiffypopjr @mattisa we'll get this into the next release.

PR #22, opened 2018-10-25
Feature/test data examples

andybluntish:
Extend the docs/examples for using test data.

- clean up code snippets in docs (consistent syntax that follow Prettier rules)
- simple example of using Faker data within test steps
- example of using Faker data as an alternative to CSV/JSON to generate records to use across multiple test iterations
- make docs on data file locations apply to all file types (not just CSV)


packages/element/docs/api/TestData.md
```diff
@@ -38,7 +38,7 @@ TestData.fromCSV<Row>('users.csv')
   .shuffle()       // Shuffle the data
 
 export default () => {
-   step('Step 1', (browser: Browser, row: Row) => {
+   step('Step 1', async (browser: Browser, row: Row) => {
```
ivanvanderbyl: We need to update this in the source code, all the docs in /api/* are generated from typescript -> markdown.

andybluntish: Good to know, thanks. This was a find/replace and included the doc comments in src files, so we should be golden.

PR #40, opened 2019-05-01
Enable external modules

ivanvanderbyl:


lachie:
FYI I implemented this directly in beta:
https://github.com/flood-io/element/blob/beta/packages/element/src/runtime/EvaluatedScript.ts#L34-L35

ivanvanderbyl:
Thanks @lachie

PR #41, opened 2019-05-01
Auto Wait

ivanvanderbyl:
Adds support for automatically applying a `wait(Until.elementIsVisible)` before each action, so you don't need to do so manually. Enabled by default.

### Usage:

Remove existing waiters from your test script:

```diff
- await browser.wait(Until.elementIsVisible(By.id("el123")))
await browser.click(By.id("el123"))
````

### Todo:

- [x] Apply to element finders
- [x] Error handling: how should we respond when a waiter times out?
- [x] Consider making the wait type configurable at global or step level
- [x] Generate docs

### CODE REVIEW

**Comment by ivanvanderbyl** [comment]
I've been thinking that it might make sense to rename this option `waitUntilVisible` to be more explicit in what it does. Alternatively we could look at other types of waits which make sense in this context, like `waitUntilPresent` and implement this as an option: `waitUntil: 'visible'`

**Comment by 90kts** [comment]
I like that idea - helps to be explicit in naming `waitUntilVisible` and code completion doesn't leave you guessing about possible args.

**Comment by ivanvanderbyl** [comment]
True, code completion would be important here. We can complete the possible valid string values by making an explicit union type for them:

```ts
export type ElementPresence = 'visible' | 'present'
```

Which would make typescript only accept those two string values.

PR #50, opened 2019-06-19
Implements File Uploads

### LINKED ISSUE #32: File uploads

There is currently no way to upload files using file inputs.

Expose an API method on the `browser` or `element` DSL for attaching a file to a file input, so we can support file uploads.

**Additional context**

This should be easy to support using [`elementHandle.uploadFile()`](https://github.com/GoogleChrome/puppeteer/blob/v1.11.0/docs/api.md#elementhandleuploadfilefilepaths)

ivanvanderbyl:
Implements file uploads for input controls.

Closes #32 

## Usage

Assuming a file was uploaded to Flood along with the script called `my-picture.png`:

```ts
let fileInput = await browser.findElement('input[type="file"]')
await fileInput.uploadFile("my-picture.png")
```


packages/element/src/page/ElementHandle.ts
```diff
@@ -108,6 +109,7 @@ function domError(
  */
 interface ScreenshotSaver {
```
lachie: This could be a bit clearer if the type were renamed to something like `FilesystemAccessor` & added an alias to `workRoot.testData` on `Browser`.

packages/element/src/page/ElementHandle.ts
```diff
@@ -108,6 +109,7 @@ function domError(
...
 	saveScreenshot(fn: (path: string) => Promise<boolean>): void
+	workRoot: WorkRoot
```
lachie: then we could replace this with
```
	testData(filename: string): string
```

packages/element/src/page/ElementHandle.ts
```diff
@@ -120,6 +122,7 @@ export class ElementHandle implements IElementHandle, Locator {
 	 * @internal
 	 */
 	private screenshotSaver: ScreenshotSaver
```
lachie:
```
private fs: FilesystemAccessor
```

packages/element/src/page/ElementHandle.ts
```diff
@@ -263,6 +267,15 @@ export class ElementHandle implements IElementHandle, Locator {
 		return handle.type(text)
 	}
 
+	/**
+	 * Sets the value of the file input these files
+	 * @param names The name of a file you uploaded with this script. Relative to the script.
+	 */
+	@wrapDescriptiveError()
+	public async uploadFile(...names: string[]): Promise<void> {
+		return this.element.uploadFile(...names.map(name => this.workRoot.testData(name)))
```
lachie: and finally
```
return this.element.uploadFile(...names.map(name => this.fs.testData(name)))
```

ivanvanderbyl: I like that. Will change

PR #51, opened 2019-06-19
Support for blocking requests to specific domains

ivanvanderbyl:
This PR adds support for blocking requests to specific domains. This can be used to prevent sending traffic to analytics providers or creating load against services which are outside of our concern.

## Usage

```ts
  blockedDomains: ['*.google-analytics.com', 'bootstrapcdn.com', '*:80'],
}
```

**Debugging:**

```shell
DEBUG=network:intercept element run script.ts
```


packages/element/src/network/Interceptor.ts
```diff
@@ -0,0 +1,36 @@
...
+
+	async detach(page: Page) {
+		page.off('request', this.requestBlocker)
+		await page.setRequestInterception(false)
+	}
+
+	private requestBlocker = (interceptedRequest: Request) => {
+		let url = new URL(interceptedRequest.url())
+
+		if (this.blockedDomains.some(domain => url.hostname.includes(domain))) {
```
90kts: wonder if `match` would work better than `includes`, as I think people might like to specify wildcards e.g. `*.cdn.com.au`

lachie: was my thought that this'd be the first thing users'd expect

ivanvanderbyl: Yeah good point. I guess we could make it a RegExp or string

ivanvanderbyl: `string.match(...)` won't really deliver what we expect here if we want to allow usual assortment of wildcards like `*.foo.com` as it's not a valid regex. I'll try the Node package https://www.npmjs.com/package/url-pattern#make-pattern-from-string

ivanvanderbyl: Actually micomatch seems like a better solution https://github.com/micromatch/micromatch

ivanvanderbyl: Done

PR #56, opened 2019-07-05
Delete .DS_Store files from macOS

wilsonmar:


90kts:
please update .gitignore instead

PR #107, opened 2020-03-02
Adds Flood Agent support package

ivanvanderbyl:


packages/flood-runner/src/ConcurrencyHack.ts
```diff
@@ -0,0 +1,16 @@
+import { readdirSync } from 'fs'
+import { join } from 'path'
+
+const lockdirPath = join(process.env.FLOOD_DATA_ROOT ?? '/data/flood', 'concurrency.lock')
```
lachie: hey, this needs to be a subdir of FLOOD_DATA_ROOT

packages/flood-runner/src/Environment.ts
```diff
@@ -0,0 +1,83 @@
...
+		SEQUENCE: this.threadID * this.gridSequenceID * (this.nodeIndex + 1),
+		FLOOD_LOAD_TEST: true,
+	}
+}
+
+export function initFromEnvironment(env: ProcessEnv = process.env): Partial<GridConfig> {
+	let root: string
+	let testDataRoot: string
+
+	if (env.NODE_ENV !== 'production') {
```
lachie: will we have to set this from flood-agent?

packages/flood-runner/src/Environment.ts
```diff
@@ -0,0 +1,83 @@
...
+		devDefaults(env)
+		const projectRoot = findRoot(__dirname)
+		root = path.join(projectRoot, 'tmp/data')
+		testDataRoot = path.join(projectRoot, 'tmp/data/flood/files')
+	} else {
+		root = '/data'
```
lachie: this isn't necessarily true anymore (outside docker)... should be set from `$FLOOD_DATA_ROOT`

packages/flood-runner/src/Environment.ts
```diff
@@ -0,0 +1,83 @@
...
+		testDataRoot = env.TEST_DATA_DIRECTORY || '/data/flood/files'
```
lachie: `$FLOOD_FILES_PATH` is probably better now

packages/flood-runner/src/ConcurrencyHack.ts
ivanvanderbyl: any pref of what it is? this is assumed to exist by Element

packages/flood-runner/src/Environment.ts
ivanvanderbyl: If it's unset it's okay, I assume this is just for smoke testing/legacy tests

PR #123, opened 2020-04-07
[bugfix] Loop Count setting does not work

### LINKED ISSUE #91: loopCount setting not working
**Describe the bug**

The loopCount setting used in TestSettings looks to have no effect on the number of iterations.

If I let it to it's default value (-1), when I run the script, it only executes once.

If I set loopCount: 10, I still get only one iteration.

The only way I can have the script running for 'X'  iterations is by setting it in the command :

element run filename --loop-count=10

**To Reproduce**

1. element generate filename
2. element run filename
2.1. Or, add loopCount: X in TestSettings and then : element run filename

**Expected behavior**

2. The script will run infinitely
2.1. The script will run for X iterations

**Desktop (please complete the following information):**

 - OS: Ubuntu [19.10]
 - Node.js version: [12.13.0]
 - Element: [e.g. 1.0.8-stable.1]

phileba:
### What does it do?
when executing the cmd: `element run file.ts`, the `loopCount` always has default value is 1 and it overrides the one which was defined in `TestSetting`. So the fix do the following:
- if the cmd does NOT included the `--loop-count` option, do not set default value is 1 and use the `loopCount` value defined in the TestSetting. If TestSetting does not define `loopCount`, the test will loop infinitely
- if the cmd included `--loop-count` option, it overrides the current one in the TestSetting.

### Ticket 
resolves #91 

### Note
It seems we do not add `.editorconfig` file so when viewing the ts file, the indent is not right.

### DIFF
--- a/packages/cli/src/cmd/run.ts
+++ b/packages/cli/src/cmd/run.ts
@@ -86,6 +86,7 @@ const cmd: CommandModule = {
 			testSettingOverrides: {},
 			persistentRunner: false,
 		}
+
 		if (args.loopCount) {
 			opts.testSettingOverrides = {
 				loopCount: args.loopCount,
--- a/packages/cli/src/cmd/run.ts
+++ b/packages/cli/src/cmd/run.ts
@@ -83,12 +83,14 @@ const cmd: CommandModule = {
 			sandbox: args.sandbox ?? true,
 
 			runEnv: initRunEnv(workRootPath, testDataPath),
-			testSettingOverrides: {
-				loopCount: args.loopCount,
-			},
+			testSettingOverrides: {},
 			persistentRunner: false,
 		}
-
+		if (args.loopCount) {
+			opts.testSettingOverrides = {
+				loopCount: args.loopCount,
+			}
+		}
 		opts.testSettingOverrides = setupDelayOverrides(args, opts.testSettingOverrides)
 
 		if (args.watch) {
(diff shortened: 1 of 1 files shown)


packages/cli/src/cmd/run.ts
```diff
@@ -83,12 +83,14 @@ const cmd: CommandModule = {
 			sandbox: args.sandbox ?? true,
 
 			runEnv: initRunEnv(workRootPath, testDataPath),
-			testSettingOverrides: {
-				loopCount: args.loopCount,
-			},
+			testSettingOverrides: {},
```
huyennbl, line 86: Should move default value to interface imo

packages/cli/src/cmd/run.ts
```diff
@@ -83,12 +83,14 @@ const cmd: CommandModule = {
...
 			persistentRunner: false,
 		}
-
+		if (args.loopCount) {
```
ivanvanderbyl, line 89: This shouldn't be required with the default removed?

phileba: This is not require, the ideal is:
- If user run the element cli likes: `element run test.ts`, the `args.loopCount` is undefined and it should not be added to `testSettingOverrides`. So the test will use the `loopCount` setting defined in TestSetting.
- If the user run `element run test.ts --loop-count=2`, the `args.loopCount` has the value so it will be added to `testSettingOverrides`, then it overrides the TestSetting defined in test file.
Is it right?

ivanvanderbyl: Fair enough.

PR #124, opened 2020-04-07
Fix screenshot on failure not working

### LINKED ISSUE #87: `screenshotOnFailure: true` does not create a screenshotOnFailure
I am running headless and can explicitly take a screenshot
```TypeScript
await browser.takeScreenshot()
````
However, the test setting ```screenshotOnFailure: true```, does not take a screenshot upon failure.

### PR DESCRIPTION
**Pull request by huyennbl**
Fix `screenshotOnFailure: true` not working 
Screenshot of errors have prefix `error_<stepName>_`
Screenshot from 2020-04-07 18-08-16
Resolves #87 

Result with example: 
Screenshot from 2020-04-13 21-41-33

### DIFF
--- a/packages/core/src/runtime/test-observers/Errors.ts
+++ b/packages/core/src/runtime/test-observers/Errors.ts
@@ -11,6 +11,10 @@ export default class ErrorObserver extends NoOpTestObserver {
 		// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
 		test.reporter.testStepError(structuredErrorToDocumentedError(err, test.script)!)
 
+		if (test.settings.screenshotOnFailure) {
+			await test.runningBrowser?.takeScreenshot()
+		}
+
 		return this.next.onStepError(test, step, err)
 	}
 }
--- a/examples/screenshot-on-failure/screenshot-on-failure.ts
+++ b/examples/screenshot-on-failure/screenshot-on-failure.ts
@@ -1,0 +1,13 @@
+import { step, By, TestSettings } from '@flood/element'
+
+export const settings: TestSettings = {
+	loopCount: 1,
+	screenshotOnFailure: true,
+}
+
+export default () => {
+	step('1. Start', async browser => {
+		await browser.visit('https://challenge.flood.io')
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: packages/core/src/runtime/test-observers/Errors.ts [hunk 1]
```diff
@@ -11,6 +11,12 @@ export default class ErrorObserver extends NoOpTestObserver {
 		// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
 		test.reporter.testStepError(structuredErrorToDocumentedError(err, test.script)!)
 
+		if (test.settings.screenshotOnFailure) {
+			// replacing whitespaces with dashes for better file name
+			const stepName = step.name.split(' ').join('-')
+			await test.runningBrowser?.takeScreenshot({ filePrefix: `error_${stepName}` })
```
ivanvanderbyl, line 17: Okay so with the prefix, this really isn't needed. Customers never interact with the actual file names, which is why we use KSUIDs as the file name. Once a screenshot is taken, it needs to be sent as a measurement to Grid so that it appears in the Flood UI at the correct time.

ivanvanderbyl: Basically what we need it to do is:

1. On error, take the screenshot
2. Ensure this is sent as a trace object so that it appears in the browser
3. Include the details of the failure in the trace to help debugging and ensure it shows as a failure on Flood

huyennbl: Hi Ivan, I reverted the prefix code. 
Noted your 2,3 but I think it's better to be in another PR/ Ticket, since I mainly try to fix issue #87 in this. I assume that those 2 items are still working well as nobody report bug about it :D

ivanvanderbyl:
Thanks @huyennbl — can you prepare an example test (put it in `./examples/`) which raises an error and captures a screenshot, to ensure this work.

huyennbl:
@ivanvanderbyl I added example, updated screenshot in PR description

ivanvanderbyl:
LGTM :D

PR #126, opened 2020-04-15
Feature/Make process.env reflect environment variables present when running the "element run"

### LINKED ISSUE #104: Make process.env reflect environment variables present when running the "element run"
**Expose process.env inside tests**
I would like to be able to access `process.env` variables in my test.ts script. Currently, the `process.env` is {} for some reason.

I would like to be able to use a tool like `env-cmd` to load env variables from file and have them accessible in the test files through `process.env`. However, this is not strictly related to a tool like `env-cmd`. It's about the general accessibility of environment variables.

**Describe alternatives you've considered**
One alternative is to use `dotenv`, which will probably work locally, but I am not sure how it will work in the flood.io cloud. I will have to test that.

**Additional context**
Is there a reason why `process.env` is not accessible / is empty within the test files? Just curious.

phileba:
### What does it do?
Provide a method to access NodeJS process env in test file

resolve #104

### DIFF
--- a/packages/core/src/runtime/types.ts
+++ b/packages/core/src/runtime/types.ts
@@ -321,8 +321,6 @@ export interface Browser {
 	switchTo(): TargetLocator
 
 	setViewport(viewport: Viewport): Promise<void>
-
-	getNodeJSProcessEnv(): NodeJS.ProcessEnv
 }
 
 /**
--- a/packages/core/src/runtime/types.ts
+++ b/packages/core/src/runtime/types.ts
@@ -321,6 +321,8 @@ export interface Browser {
 	switchTo(): TargetLocator
 
 	setViewport(viewport: Viewport): Promise<void>
+
+	getNodeJSProcessEnv(): NodeJS.ProcessEnv
 }
 
 /**
--- a/global.d.ts
+++ b/global.d.ts
@@ -11,5 +11,3 @@ declare module 'comment-parser'
 declare module 'update-check'
 
 declare module 'yeoman-generator'
-
-declare let processEnv: NodeJS.Process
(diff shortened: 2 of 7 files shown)


packages/core/src/runtime/types.ts
```diff
@@ -321,6 +321,8 @@ export interface Browser {
 	switchTo(): TargetLocator
 
 	setViewport(viewport: Viewport): Promise<void>
+
+	getNodeJSProcessEnv(): NodeJS.ProcessEnv
```
ivanvanderbyl, line 325: I don't think this is a good solution because this isn't how you would expect to access the environment in Node.

The VM (created in EvaluatedScript) has an environment entrypoint which we can selectively include the properties from process.env

ivanvanderbyl:
```
Object.keys(process.env)
```

phileba: @ivanvanderbyl  yep I got your point but if we include all properties of `process.env` into the sandbox, actually we can access from the test script by calling only the name of the key like:
image

ivanvanderbyl: I think we should make this as natural and consistent with node as possible, otherwise we will create more inertia to using it and more things we need to document.

So what about:

```
sandbox: {
  process: { env: process.env },
},
```

If that's legal, not sure.

phileba: I tried this but it does not work, that's why I thought about 'processEnv`

phileba: The process.env set to {} in the sandbox at https://github.com/patriksimek/vm2/blob/2049e4d34d35ff4fc9cf83390571c7b06eeb2e77/lib/sandbox.js#L388-L395, have an issue at https://github.com/patriksimek/vm2/issues/282

PR #136, opened 2020-04-21
Feature/Log step timings when running locally

### LINKED ISSUE #89: Log step timings when running locally
Basic idea is to log the timing of each step when completed locally. At the moment, we only show the aggregate time for all steps at the end of the loop.

phileba:
### What does it do?
Log the timing of each step when completed locally.

image

resolve #89

### DIFF
--- a/packages/cli/src/utils/ConsoleReporter.ts
+++ b/packages/cli/src/utils/ConsoleReporter.ts
@@ -13,7 +13,7 @@ const debug = require('debug')('element-cli:console-reporter')
 export class ConsoleReporter implements IReporter {
 	public responseCode: string
 	public stepName: string
-	public startTime = 0
+	private startTime = 0
 	constructor(private logger: Logger, private verbose: boolean) {}
 
 	reset(step: string): void {}
--- a/packages/cli/src/utils/ConsoleReporter.ts
+++ b/packages/cli/src/utils/ConsoleReporter.ts
@@ -46,7 +46,7 @@ export class ConsoleReporter implements IReporter {
 				this.logger.info(`===> Step '${label}'`)
 				break
 			case TestEvent.AfterStep:
-				this.logger.info(`---> Step '${label}' finished in ${timing}ms`)
+				this.logger.info(`---> Step '${label}' finished in ${timing?.toLocaleString()}ms`)
 				break
 			case TestEvent.StepSkipped:
 				this.logger.info(`---- Step '${label}' skipped`)
--- a/packages/cli/src/utils/ConsoleReporter.ts
+++ b/packages/cli/src/utils/ConsoleReporter.ts
@@ -5,6 +5,7 @@ import {
 	CompoundMeasurement,
 	MeasurementKind,
 	TestScriptError,
+	Timing,
 } from '@flood/element-core'
 import { Logger } from 'winston'
 import chalk from 'chalk'
@@ -13,7 +14,8 @@ const debug = require('debug')('element-cli:console-reporter')
 export class ConsoleReporter implements IReporter {
 	public responseCode: string
 	public stepName: string
-	private startTime = 0
+	private t: Timing = new Timing()
+
 	constructor(private logger: Logger, private verbose: boolean) {}
 
 	reset(step: string): void {}
@@ -42,13 +44,16 @@ export class ConsoleReporter implements IReporter {
 				this.logger.info(`---> ${label}()`)
 				break
 			case TestEvent.BeforeStep:
-				this.startTime = new Date().valueOf()
 				this.logger.info('')
 				this.logger.info(`===> Step '${label}'`)
+				this.t.start('step')
 				break
 			case TestEvent.AfterStep:
+				this.t.end('step')
 				this.logger.info(
-					`---> Step '${label}' finished in ${new Date().valueOf() - this.startTime}ms`,
+					`---> Step '${label}' finished in ${this.t.getDurationWithoutThinkTimeForSegment(
+						'step',
+					)}ms`,
 				)
 				break
 			case TestEvent.StepSkipped:
--- a/packages/cli/src/utils/ConsoleReporter.ts
+++ b/packages/cli/src/utils/ConsoleReporter.ts
@@ -13,7 +13,8 @@ const debug = require('debug')('element-cli:console-reporter')
 export class ConsoleReporter implements IReporter {
 	public responseCode: string
 	public stepName: string
-
+	public startTime = 0
+	public endTime = 0
 	constructor(private logger: Logger, private verbose: boolean) {}
 
 	reset(step: string): void {}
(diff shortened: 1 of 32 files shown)


packages/cli/src/utils/ConsoleReporter.ts
```diff
@@ -13,7 +13,7 @@ const debug = require('debug')('element-cli:console-reporter')
 export class ConsoleReporter implements IReporter {
 	public responseCode: string
 	public stepName: string
-
+	public startTime = 0
```
huyennbl, line 16: `private`

packages/cli/src/utils/ConsoleReporter.ts
```diff
@@ -42,11 +42,14 @@ export class ConsoleReporter implements IReporter {
 				this.logger.info(`---> ${label}()`)
 				break
 			case TestEvent.BeforeStep:
+				this.startTime = new Date().valueOf()
```
huyennbl, line 45: this.startTime = performance.now()

packages/cli/src/utils/ConsoleReporter.ts
```diff
@@ -42,11 +42,14 @@ export class ConsoleReporter implements IReporter {
...
 				this.logger.info('')
 				this.logger.info(`===> Step '${label}'`)
 				break
 			case TestEvent.AfterStep:
-				this.logger.info(`---> Step '${label}' finished`)
+				this.logger.info(
+					`---> Step '${label}' finished in ${new Date().valueOf() - this.startTime}ms`,
```
huyennbl, line 51: performance.now() - this.startTime

packages/cli/src/utils/ConsoleReporter.ts
huyennbl: (not sure if this still works with private access modifier, if it doesn't please ignore this comment)

packages/cli/src/utils/ConsoleReporter.ts
phileba: `performance` cannot be used as a native module, you need to import `perf_hooks` as the following: 
`const { performance } = require('perf_hooks');` before using. There is no difference in this use case so I prefer to use the native module

packages/cli/src/utils/ConsoleReporter.ts
phileba: same as comment above

packages/cli/src/utils/ConsoleReporter.ts
ivanvanderbyl, line 45: Okay, sooo, this isn't really correct.. step timings as wall clock time are only one way we measure things, we have a few different timing measurements implemented, which you should need to hook into.

phileba: okay, so i think we should move the implementation of class `Timing` in `packages/flood-runner/src/test-observers/Timing.ts`
```
class Timing {...}
````
 to core instead of runner, keep the rest of this file. So we can use those timing measurements as common. How about your ideal @ivanvanderbyl ?

Comment (ivanvanderbyl) [reply packages/cli/src/utils/ConsoleReporter.ts:45]: That sounds good. IIRC the dependency graph for those two packages is cli -> runner -> core, so it should just work.

File: packages/cli/src/utils/ConsoleReporter.ts [hunk 4]
```diff
@@ -44,9 +46,15 @@ export class ConsoleReporter implements IReporter {
 			case TestEvent.BeforeStep:
 				this.logger.info('')
 				this.logger.info(`===> Step '${label}'`)
+				this.t.start('step')
 				break
 			case TestEvent.AfterStep:
-				this.logger.info(`---> Step '${label}' finished`)
+				this.t.end('step')
+				this.logger.info(
+					`---> Step '${label}' finished in ${this.t.getDurationWithoutThinkTimeForSegment(
```
ivanvanderbyl, line 54: This is a good start, however, we need it to select which type of timing it uses based on the settings in the test script. See TestSettings.

phileba: @ivanvanderbyl you mean those kinds of `"page" | "network" | "step" | "stepWithThinkTime"`? so I think we need to move the function `getResponseTimeMeasurement` to the core then we can use it as common also.

ivanvanderbyl: That makes sense, do it!

packages/cli/src/utils/ConsoleReporter.ts
```diff
@@ -13,6 +14,7 @@ const debug = require('debug')('element-cli:console-reporter')
 export class ConsoleReporter implements IReporter {
 	public responseCode: string
 	public stepName: string
+	private t: Timing = new Timing()
```
ivanvanderbyl, line 17: Please use full variable names in wide scope.

packages/cli/src/utils/ConsoleReporter.ts
phileba: @ivanvanderbyl, I just move the `getResponseTimeMeasurement(ResponseTiming, NetworkRecorder)` to the core and we can use it as a common function but there is a problem: if we call this function in CLI in somewhere likes `ConsoleReporter.ts`, how to get the params of `NetworkRecorder`?
So the ideal is:
- keep the `getResponseTimeMeasurement` in class `TimingObserver` and write new function likes
```
async getMeasurementTime(responseTiming: ResponseTiming): Promise<number> {
	await this.syncNetworkRecorder()
	const { networkRecorder } = this.ctx
	return this.getResponseTimeMeasurement(responseTiming, networkRecorder)
}
```
- move this from runner to the core
- modify function `runWithCancellation` in `Test.ts` of the core by adding
```
const ctx = new Context()
const timingObserve = new TimingObserver(ctx,
	new NetworkRecordingTestObserver(ctx, new NullTestObserver()),
)
...
//hook to before test for timingObserve
await timingObserve.before(this)
```
- in `Test.ts` get timing by calling `timingObserve.getMeasurementTime(this.settings.responseTimeMeasurement)`, we can call this function in CLI also.

_=> Not sure it right but always get 0 in the result._

ivanvanderbyl: I think I follow what you're saying, let's try it out

phileba: pushed 👌 @ivanvanderbyl

ivanvanderbyl: 👀

phileba: it builds and runs success but `yarn test` fail at `core/src/runtime/test-observers/Timing.ts`, not sure why?

phileba:
> It looks good and works locally for me.
> 
> One thing, can you correct the formatting of the response time number to be correctly formatted:
> 
> It should read as `2,345 ms` not `2345ms`

resolved

Issue #133, opened 2020-04-21
Support a configuration file for reusing a config across tests within a project/repo


ivanvanderbyl:
Similar to how jest does it. See https://github.com/flood-io/element/projects/1#card-31266795

Features:

- Support configuring element from a file e.g. `element.config.js`
- Ability to specify all standard CLI flags and options
- Ability to specify a path to element tests e.g. `testPathMatch: "tests/performance/*.perf..ts"`

```
element run
```

Load the config, and run all tests.

phileba:
I think we should allow the user to decide to load a config file as a CLI flag like:
`element run --config-file='./element.config.js'`
How do you think @ivanvanderbyl?

ivanvanderbyl:
Prior art: https://github.com/facebook/jest/tree/master/packages/jest-config

I'm thinking something along the lines of `element-config`..

ivanvanderbyl:
Initialize new config:

```
element --init // generates config file
```

PR #137, opened 2020-04-21
Allow user to specify a return code when the test fails

### LINKED ISSUE #30: Exit status of `element run test.ts` to reflect test passing/failing
Feature request. Use case: before the full load test runs in CI (like Jenkins) - validate that script is still functional - without spinning grid or using hours on flood.io

return of `element run test.ts` to be not 0 if test failed. Right now regardless if test is passing or failing I am seeing "0" return code.

**Describe alternatives you've considered**
Some magic script to parse output?

**Additional context**
None

HieuLeKMS:
User can pass in the option **fail-status-code** to the command `element run test.ts` to be the exit code when the test fails, otherwise exit code is 0.

resolve #30

### DIFF
--- a/packages/cli/src/cmd/run.ts
+++ b/packages/cli/src/cmd/run.ts
@@ -32,6 +32,7 @@ interface RunArguments extends Arguments {
 	slowMo?: number
 	'work-root'?: string
 	'test-data-root'?: string
+	'fail-status-code'?: number
 }
 
 function setupDelayOverrides(args: RunArguments, testSettingOverrides: TestSettings) {
@@ -85,6 +86,7 @@ const cmd: CommandModule = {
 			runEnv: initRunEnv(workRootPath, testDataPath),
 			testSettingOverrides: {},
 			persistentRunner: false,
+			failStatusCode: args['fail-status-code'],
 		}
 
 		if (args.loopCount) {
@@ -187,6 +189,9 @@ const cmd: CommandModule = {
 			})
 			.option('verbose', {
 				describe: 'Verbose mode',
+			})
+			.option('fail-status-code', {
+				describe: 'Exit code when the test fails',
 			})
 			.positional('file', {
 				describe: 'the test script to run',
--- a/packages/cli/src/cmd/run.ts
+++ b/packages/cli/src/cmd/run.ts
@@ -192,6 +192,7 @@ const cmd: CommandModule = {
 			})
 			.option('fail-status-code', {
 				describe: 'Exit code when the test fails',
+				type: 'number',
 			})
 			.positional('file', {
 				describe: 'the test script to run',
--- a/packages/cli/src/cmd/run.ts
+++ b/packages/cli/src/cmd/run.ts
@@ -1,7 +1,6 @@
 /* eslint-disable @typescript-eslint/no-use-before-define */
 import {
 	runCommandLine,
-	runUntilExit,
 	ElementOptions,
 	WorkRoot,
 	FloodProcessEnv,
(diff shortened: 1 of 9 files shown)


packages/cli/src/cmd/run.ts
```diff
@@ -32,6 +32,7 @@ interface RunArguments extends Arguments {
 	slowMo?: number
 	'work-root'?: string
 	'test-data-root'?: string
+	'fail-status-code'?: number
```
huyennbl, line 35: Just suggestion, how about we use this one as non-nullable var, with default value of 0 :eyes:

HieuLeKMS: `RunArguments` is the type for the options of the command `element run`, so the value for `fail-status-code` can be `undefined` as it might not be passed to the command

ivanvanderbyl: I think a default makes sense, but fail should be exit 1.

packages/core/src/Element.ts
```diff
@@ -27,12 +27,13 @@ export interface ElementOptions {
...
+	failStatusCode?: number
 }
 
-export function runUntilExit(fn: () => Promise<void>) {
+export function runUntilExit(fn: () => Promise<number>) {
 	fn()
-		.then(() => {
-			process.exit(0)
+		.then((code: number) => {
+			process.exit(code)
```
ivanvanderbyl, line 36: If this is failStatusCode shouldn't this only get called within catch?

packages/cli/src/cmd/run.ts
HieuLeKMS: So default value is 1 for `fail-status-code` then

packages/core/src/Element.ts
HieuLeKMS: You mean we throw the `failStatusCode` as an error to catch is here, instead of return it as a result of running tests?

packages/cli/src/cmd/run.ts
HieuLeKMS: But I remember you said that if user doesn't specify `fail-status-code`, it would exit 0 regardless of the result. So shouldn't its default value is 0? @ivanvanderbyl

ivanvanderbyl: I did say that, and we should check that this will work in Grid, I. expect it will.

packages/core/src/Element.ts
ivanvanderbyl: If a test completes without error, we should always exit `0`, if a test fails or throws and error or for whatever reason is not a clean exit, it should exit with `failStatusCode`

packages/cli/src/cmd/run.ts
HieuLeKMS: @ivanvanderbyl  by "work in Grid" you mean making this work with the command `element agent start`? Because I don't know whether there's a way for user to specify `fail-status-code` when they upload and run tests on Flood

ivanvanderbyl: Correct, currently there isn't a way to do that, however, in the future when we have #133 implemented, this will be possible. Don't worry about this now, as long as it works locally.

packages/core/src/Element.ts
ivanvanderbyl: Still waiting on this.

HieuLeKMS: I removed the function `runUntilExit`, and handled exit logic in `runCommandLine`. Since `failStatusCode` and other `ElementOptions` is passed into `runCommandLine`, I think it's better to handle exit there

PR #138, opened 2020-04-23
Feature/Send keys combination

### LINKED ISSUE #13: sendKeys Key combinations (incl cross-platform)
I'd like to be able to do key combos using `browser.sendKeys()` and the `Key` enum to simulate typing on a keyboard.
Additionally I need to send different key combinations depending on the platform the test script will run on.

Command-Key combos might not be possible on mac, but there are still plenty of web-apps which handle keyboard input directly for key combos.
https://github.com/GoogleChrome/puppeteer/issues/1313

When developing the test script using Element on my Mac I'd copy something with CMD-C, but when the script runs on flood.io it'll be running under linux, where the equivalent combo is CTRL-C. This point may be moot depending on puppeteer#1313

A simple API would be to allow grouping `sendKeys` inputs into arrays: `browser.sendKeys( [Key.CONTROL, "A"] )` would press CTRL-A

Not all keys [listed](https://github.com/GoogleChrome/puppeteer/blob/v1.8.0/lib/USKeyboardLayout.js) are included in the Key enum

phileba:
### What does it do?
Allow user to send key combinations likes:
```
browser.sendKeyCombinations(Key.SHIFT, 'KeyA')
```
### Anything to note?
```
Puppeteer doesn't emulate chrome-level input, so there are some combination keys does not work
On Win and Linux:
Ctrl-A, Ctrl-C, Ctrl-V, Ctrl-F, Ctrl-S...
On OS X:
 ⌘-A, ⌘-C, ⌘-V, ⌘-F, ⌘-S...
```
resolve #13


phileba:
> Can you please add some tests for these? There should be a test page for keyboard inputs in the dogfood repo

okay sure

packages/core/src/runtime/Browser.spec.ts
```diff
@@ -165,4 +166,18 @@ describe('Browser', () => {
...
+
+	describe('send key combination', () => {
+		test('can send combination keys', async () => {
+			const browser = new Browser(workRoot, puppeteer, DEFAULT_SETTINGS)
+			const url = await serve('combination_keys.html')
+			await browser.visit(url)
+			const input = await browser.findElement(By.id('text'))
+			await input.focus()
+			await browser.sendKeyCombinations('KeyA')
+			await browser.sendKeyCombinations('Shift', 'KeyA')
```
ivanvanderbyl, line 178: What is `'KeyA'`?

extern/dogfood/dockerfiles/watirspec/html/combination_keys.html
```diff
@@ -0,0 +1,9 @@
+<!DOCTYPE html>
```
ivanvanderbyl, line 1: This should be a submodule, so this change shouldn't be in this repo. Cna you investigate if A) there's not already a test for keyboard combinations, and B) why this is showing up here

packages/core/src/runtime/Browser.spec.ts
```diff
@@ -5,6 +5,7 @@ import { Browser } from './Browser'
 import { Until } from '../page/Until'
 import { By } from '../page/By'
 import { DEFAULT_SETTINGS } from './Settings'
+import * as assert from 'assert'
```
ivanvanderbyl, line 8: Shouldn't this be a default import?

packages/core/src/runtime/Browser.spec.ts
phileba: yep, you can have a look at this https://pptr.dev/#?product=Puppeteer&version=v3.0.2&show=api-class-keyboard, on using combination keys, we need to input the key, details at https://github.com/puppeteer/puppeteer/blob/master/src/USKeyboardLayout.ts

packages/core/src/runtime/Browser.spec.ts
phileba: No, it did not imported as default, we can change this by `expect(...).toBe(...)`

extern/dogfood/dockerfiles/watirspec/html/combination_keys.html
phileba: so it seems there is no test file not only for keyboard combination but also for send key. So I wanna create an Html file that only has a text field then navigate to and test the combination key.

ivanvanderbyl: Okay fair enough, thanks for checking. I assumed there would be as that's the test case used for another QA tool. I guess keyboard interaction isn't very common in older tools.

phileba: okay, so just leave this test file here.

packages/core/src/page/Enums.ts
```diff
@@ -159,3 +159,54 @@ export const userAgents = {
 	operaWindows: 'Opera/9.80 (Windows NT 6.1; U; en) Presto/2.10.229 Version/11.62',
 	robot: 'I AM ROBOT',
 }
+
+export const KeyDefinitions = {
```
ivanvanderbyl, line 163: Why do we vendor this instead os using the keys declared by puppeteer?

phileba: I know you will ask this question, this is cause the puppeteer does not export the KeyDefinitions, you can see this https://github.com/puppeteer/puppeteer/blob/master/src/api.js

ivanvanderbyl: Ah okay, fair enough, makes sense then.[](url)

PR #148, opened 2020-05-05
Feature/Ability to mark a step as "once off"

### LINKED ISSUE #69: Ability to mark a step as "once off"

phileba:
### What does it do
Provide an option to mark the step as "once-off", the idea is:
```
step.once('Test: 1', async browser => {
	console.log('step 1')
})
```
resolve #69


ivanvanderbyl:
I think this requires further discussion because a change like this should be more than just a configuration option.

I'd like to suggest something like this (inspired by Jest/Jasmine test):

```
step.once("Login", async browser => {})
```

phileba:
@ivanvanderbyl yep it's a good idea but I'm not sure how to add something likes `step.once`. Do we have any document or sample?

ivanvanderbyl:
Here's an example of how it is done in Jest https://github.com/facebook/jest/blob/master/packages/jest-circus/src/index.ts#L21-L38

ivanvanderbyl:
What they're doing is defining skip (could be once, repeat, recover etc) as a method on the describe function. When called it just hoists the callback function to the describe block passing a flag similar to yours which sets the state.

It's purely syntactic sugar, but makes for a better tester experience.

packages/core/src/runtime/EvaluatedScript.ts
```diff
@@ -165,7 +172,16 @@ export class EvaluatedScript implements TestScriptErrorMapper, EvaluatedScriptLi
...
-			steps.push({ fn, name, stepOptions })
+
+			steps.push({ fn, name, stepOptions, type })
+		}
+
+		function captureStepNormal(...args: any[]) {
+			captureStep(args, StepType.STEP)
+		}
+
+		function captureStepOnce(...args: any[]) {
```
ivanvanderbyl, line 183: I reckon just inline these as expressions below, this just reads really verbosely.

packages/core/tests/fixtures/vm-features.ts
```diff
@@ -17,11 +17,15 @@ export default () => {
...
 
 	// First step
-	step('Step 1', { waitTimeout: 60 }, async driver => {
-		await driver.click('#element1')
-		await driver.click('#element2')
-		await driver.click('#element3')
-	})
+	step(
+		'Step 1',
+		async driver => {
```
ivanvanderbyl, line 22: This looks like a breaking change

ivanvanderbyl: Why did you change the order of the arguments?

packages/core/src/runtime/Step.ts
```diff
@@ -8,31 +8,26 @@ import { ElementPresence } from './Settings'
...
  *
  *   step("Step 3", async (browser: Browser) => {})
  * }
  * ```
  *
  * @param name Step Name
  * @param fn Actual implementation of step
+ * @param options step options
  */
-export declare function step<T>(name: string, fn: StepFunction<T>): void
```
ivanvanderbyl, line 24: These declarations aren't correct. The Jest example I showed you should lead to a solution here.

packages/core/tests/fixtures/vm-features.ts
phileba: because I change the declaration of the step so I change the order of the arguments, likes:
```
export const step = (() => {
	function step<T>(name: string, fn: StepFunction<T>, options?: StepOptions): void {}
	function once<T>(name: string, fn: StepFunction<T>, options?: StepOptions): void {}
	step.once = once
	return step
})()
```

ivanvanderbyl:
Can you merge this into #150

phileba:
> Can you merge this into #150

sure (y)

PR #149, opened 2020-05-12
Handle multiple Tabs/Windows

### LINKED ISSUE #47: Handle multiple Tabs/Windows
Add support for switching tabs or windows in much the same way we do with Frames, using the `TargetLocator`.

HieuLeKMS:
Add some properties and methods to handle multiple tabs

- **Browser.pages**: get currently open tabs in browser
- **Browser.waitForNewPage()**: wait for a tab to be opened in browser
- **TargetLocator.page()**: switch to another tab in the browser

resolve #47


packages/core/src/runtime/Browser.spec.ts
```diff
@@ -195,4 +195,27 @@ describe('Browser', () => {
...
 		})
 	})
+
+	test('Multiple pages handling', async () => {
+		const browser = new Browser(workRoot, puppeteer, DEFAULT_SETTINGS)
+		const url = await serve('page_1.html')
+
+		await browser.visit(url)
+		await browser.click(By.tagName('a'))
+		await browser.waitForNewPage()
```
ivanvanderbyl, line 205: It'd be nice if this returned the new page:

```
let newPage = await browser.waitForNewPage()
await browser.switchTo().page(newPage)
```

However, I expect we might have a race condition here: clicking the link then waiting. The click will happen, and possibly return, before we wait. This is also a problem with alert boxes, so we need to think this over a bit

ivanvanderbyl: Possibly the easiest would be to return the most recently opened tab when waiting, but this would only work once 🤔

ivanvanderbyl: _Thinking out loud here:_

If we declared what we want to do with the new page immediately, we could probably assume that this code block should run within the new tab, which could be assigned before that code block is reached (because of await click completing), and discarded at the end of the block.

```ts
await browser.click("a.opener")
// new page is assigned here through event handler
await useNewPage(async browser => {
  await browser.click(...)
})
// new page is discarded here

// back in original page context
await browser.select(...)
```

HieuLeKMS: I don't understand the part about race condition you mentioned above. What currently being done here is waiting for the click to be done first, then waiting for the new tab to appear, so there should be no race condition

HieuLeKMS: You mean by using `useNewPage` we don't need to `waitForNewPage`?

But is that a new instance of browser inside `useNewPage`? And would the block inside `useNewPage` run in parallel with the code right after `useNewPage`?

HieuLeKMS: @ivanvanderbyl Could you give me more details on your idea? Sorry, I cannot fully understand what you mean

ivanvanderbyl: Let's try your way for now, my idea isn't fully baked yet.

PR #150, opened 2020-05-15
Feature/Allow to add step condition

### LINKED ISSUE #67: Ability to mark a step as conditional on a value extracted from another step

phileba:
### What does it do
Provide an step condition to run the step, the idea is:
```
step.if(elementIsVisible('loginBtn'), 'Test: 1', async browser => {
	console.log('step 1')
})
```
resolve #67


packages/core/src/runtime/EvaluatedScript.ts
```diff
@@ -165,29 +173,52 @@ export class EvaluatedScript implements TestScriptErrorMapper, EvaluatedScriptLi
...
+			captureStep([name, { ...option, once: true }, fn])
+		}
+
+		const stepIf = async (conditionFn: ConditionFn, name: string, ...optionsOrFn: any[]) => {
+			const [option, fn] = extractOptionsAndCallback(optionsOrFn)
+			captureStep([name, { ...option, ifFn: conditionFn }, fn])
+		}
+
+		const step = (() => {
+			const step: any = stepNomal
```
ivanvanderbyl, line 216: I wrote you some types for this, please keep avoid using `any` — it removes all the benefits of typing, and causes bugs in refactoring later.

packages/core/src/runtime/Step.ts
```diff
@@ -20,48 +20,64 @@ import { ElementPresence } from './Settings'
...
+	 * Creates a conditional step, which will only run if the preceeding predicate returns false
+	 */
+	unless: StepConditionBase
+}
+
+export type StepDefinition = (name: string, fn: TestFn) => Promise<any>
+export type TestFn = (this: void, browser: Browser) => Promise<any>
+export type ConditionFn = (this: void, browser: Browser) => boolean | Promise<boolean>
+export type StepOptions = {
+	pending?: boolean
```
ivanvanderbyl, line 63: If a step is pending, we should generate a blank test fn which just logs `(Pending) ${stepName}`

packages/core/src/runtime/Test.ts
```diff
@@ -159,6 +159,15 @@ export default class Test implements ITest {
 
 			debug('running steps')
 			for (const step of this.steps) {
+				const { once, predicate } = step.options
+				if (once && iteration > 1) {
+					continue
+				}
+				if (predicate) {
+					const ifCondition = await predicate.call(null, browser)
+					if (!ifCondition) continue
```
ivanvanderbyl, line 168: It'd be good to have some debug log here when in development (using element locally) to indicate that the step was skipped because of the condition failing

phileba:
> What do you think about adding `step.skip` for completeness? We'd still log this, but it essentially wouldn't do anything.

sure, will update soon

packages/core/src/runtime/Test.ts
```diff
@@ -159,6 +159,26 @@ export default class Test implements ITest {
...
+				}
+				if (once && iteration > 1) {
+					continue
+				}
+				if (skip) {
+					console.log(`Skip test ${step.name}`)
+					continue
+				}
+				if (predicate) {
+					const ifCondition = await predicate.call(null, browser)
```
ivanvanderbyl, line 175: Something we need to think about here: how are errors handled inside the predicate? If this promise rejects because of a timeout error, or just a throw inside the predicate, this should return false.

ivanvanderbyl, line 175:
```suggestion
					const ifCondition = await predicate.call(null, browser).catch(err => { log(err); return false }
```

phileba: yep I think so 👌

phileba: I think this one will not work, it should be
```
if (predicate) {
	let ifCondition = false
	try {
		ifCondition = await predicate.call(null, browser)
	} catch (err) {
		console.log(err.message)
	}
	if (!ifCondition) {
		debug('condition failling')
		continue
	}
}
```
so let try with this test
```
step.if(
	() => {
		throw Error('Error')
	},
	'Test unless',
	async browser => {
		console.log('Test unless runs with condition false')
	},
)
```
the error with be catched and the step stopped as expected

phileba:
@ivanvanderbyl when I pull the new project to my local machine then install the dependencies, I always get some errors with `node-gyp` and `fsevents`. The element can run the test but I'm not sure behind the sense, there is no effect on the whole application. So. I remove the current `yarn.lock` then reinstall to generate the new one. build and test are green, test script runs success as well. Can we consider to update the latest one to the current one?
[log.txt](https://github.com/flood-io/element/files/4655491/log.txt)

PR #160, opened 2020-05-20
feature/add step recovery

### LINKED ISSUE #70: Ability to mark a step as a recovery step
The idea behind this would be to declare a step which runs when any step fails, as a sort of catch-all. This could be used when the state of a page is not as expected, or authentication fails requiring starting at step 1 again.

```ts
step.recovery(async browser => {
  ...
})

// or with info

type RecoveryInfo = {
  failedStep: string
  reason: FailureReason
}

enum FailureReason {
  Timeout
  ElementNotFound
  PageCrashed
  // etc
}

step.recovery(info: RecoveryInfo, async browser => {
  ...
})
```

**Unknowns:**

- Need to think about how to redirect to the correct step afterward
- Should recovery be associated with a single step?

phileba:
### What does it do?
The idea behind this would be to declare a step which runs when any step fails, as a sort of catch-all. This could be used when the state of a page is not as expected, or authentication fails requiring starting at step 1 again. We could take this further and do something like this:

```ts
// Step 1 recovery
	step.recovery('Step 1', async browser => {
		//do stuff
		return RecoveryOption.RETRY
	})

	step('Step 1', async browser => {
		//do stuff
	})
```
resolve #70

### DIFF
--- a/examples/condition/test.ts
+++ b/examples/condition/test.ts
@@ -23,7 +23,7 @@ export default () => {
 		},
 	)
 	step.skip('Test skip', async browser => {
-		console.log('Test skip')
+		console.log('This step will not be run')
 	})
 	step('Step pending')
 
--- a/examples/condition/test.ts
+++ b/examples/condition/test.ts
@@ -1,18 +1,13 @@
-import { step, TestSettings } from '@flood/element'
+import { step, TestSettings, RecoveryOption } from '@flood/element'
 
 export const settings: TestSettings = {
-	loopCount: 2,
+	loopCount: 1,
 }
 
 export default () => {
-	step('Test normal', async browser => {
-		console.log('Step normal is still runs')
-	})
-
 	step.once('Test once', async browser => {
 		console.log('Step once runs 1 time')
 	})
-
 	step.if(
 		() => true,
 		'Test if',
@@ -20,7 +15,6 @@ export default () => {
 			console.log('Step if runs with condition true')
 		},
 	)
-
 	step.unless(
 		() => false,
 		'Test unless',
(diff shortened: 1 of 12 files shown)


examples/condition/test.ts
```diff
@@ -1,37 +1,76 @@
...
+	})
+
+	step('Step 3', async browser => {
+		console.log('Throw error step 3')
+		throw Error('Fail')
+	})
+
+	step('Step 4', async browser => {
+		console.log('Done step 4')
+	})
```
HieuLeKMS, line 75: This is just a normal step. Why do we need it here?

packages/core/src/runtime/Step.ts
```diff
@@ -100,6 +113,23 @@ export function extractOptionsAndCallback(args: any[]): [Partial<StepOptions>, T
...
+export type StepRecoveryObject = {
+	[name: string]: Step
+}
+
+/**
+ * The `RecoveryOption` represents action which we will do after the recovery step has finished
+ * ```
+ * RETRY: retry the step failed
+ * RESTART: re-run the first step
+ * ERROR: throw an error
```
HieuLeKMS, line 125: Should this be better?
```suggestion
 * Possible actions to do after a step failed
 * ```
 * RETRY: retry the failed step
 * RESTART: re-run the step
 * ERROR: throw an error
```

examples/condition/test.ts
```diff
@@ -1,37 +1,76 @@
...
 	step.unless(
 		() => false,
 		'Test unless',
 		async browser => {
 			console.log('Test unless runs with condition false')
 		},
 	)
-
 	step.skip('Test skip', async browser => {
 		console.log('Test skip')
```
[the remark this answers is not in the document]
phileba: yep, that's a good point

examples/condition/test.ts
```diff
@@ -1,37 +1,76 @@
...
 	})
-
 	step('Step pending')
```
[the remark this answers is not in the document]
phileba: - `skip`: When you are maintaining a large codebase, you may sometimes find a test that is temporarily broken for some reason. If you want to skip running this test, but you don't want to delete this code, you can use `test.skip` to specify some tests to skip.
- `step pending`: I think there are some tests that we just have only the ideal but do not implement, we can write down for the later use.

examples/condition/test.ts
phileba: The recovery step 3 throws an error and the test will stop immediately, so the `Step 4` here to let the user know this will not be executed.

packages/core/src/runtime/Step.ts
phileba: it's good but I think it does not clear as the original one.

packages/core/src/runtime/Test.ts
```diff
@@ -164,23 +169,50 @@ export default class Test implements ITest {
...
+			}
+
+			const callRecovery = async (step: Step): Promise<boolean> => {
+				const recoveryStep = this.recoverySteps[step.name]
+				delete this.recoverySteps[step.name] // step recovery should be run 1 time
+				if (!recoveryStep) return false
+				try {
+					const result = await recoveryStep.fn.call(null, browser)
+					if (result === RecoveryOption.ERROR) return false
+					if (result === RecoveryOption.RESTART) this.stepCount = 0
```
ivanvanderbyl, line 202: This is probably not the best way to restart, because it will skip all the cleanup tasks. We should instead just exit this loop and let all that run, assuming it will restart if loopCount > 1

packages/core/src/runtime/Step.ts
```diff
@@ -100,6 +113,23 @@ export function extractOptionsAndCallback(args: any[]): [Partial<StepOptions>, T
...
+
+/**
+ * The `RecoveryOption` represents an action which we will do after the recovery step has finished
+ * ```
+ * RETRY: retry the failed step
+ * RESTART: re-run the first step
+ * ERROR: throw an error
+ * ```
+ */
+export enum RecoveryOption {
```
ivanvanderbyl, line 128: This is nice, but doesn't read very well. Maybe instead we can call it `RecoverWith`.

```ts
return RecoverWith.RESTART
```

ivanvanderbyl:
As discussed we will add:

- max recovery count option at test level and step level
- provide 3 keywords for recovery: retry, continue, restart

packages/core/src/runtime/Test.ts
phileba: so if the loop is running at iteration = 1 and we decide to restart in `recoveryStep`, so we will exit the loop and continue with iteration 1 again

PR #166, opened 2020-05-28
fix: Unable to use the command line: generate <file.ts> #165

### LINKED ISSUE #165: Unable to use the command line: generate <file.ts>
**Describe the bug**
Unable to use the command line: generate <file.ts> 

**To Reproduce**
`element generate script.ts`

**Expected behavior**
The test script should be generated success

occurs on element v1.2.3

ai-flood:
resolve #165

### DIFF
--- a/packages/cli/src/cmd/generate.ts
+++ b/packages/cli/src/cmd/generate.ts
@@ -1,14 +1,19 @@
 import { Argv, Arguments, CommandModule } from 'yargs'
 
+interface CommandArgs extends Arguments {
+	file: string
+}
+
 const cmd: CommandModule = {
 	command: 'generate <file> [options]',
 	describe: 'Generate a basic test script from a template',
 
-	async handler(args: Arguments) {
+	async handler(args: CommandArgs) {
 		const { default: yeomanEnv } = await import('yeoman-environment')
 		const env = yeomanEnv.createEnv()
-		env.register(require.resolve('../generator/test-script'), 'test-script')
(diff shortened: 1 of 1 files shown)


packages/cli/src/cmd/generate.ts
```diff
@@ -1,14 +1,19 @@
...
 	command: 'generate <file> [options]',
 	describe: 'Generate a basic test script from a template',
 
-	async handler(args: Arguments) {
+	async handler(args: CommandArgs) {
 		const { default: yeomanEnv } = await import('yeoman-environment')
 		const env = yeomanEnv.createEnv()
-		env.register(require.resolve('../generator/test-script'), 'test-script')
-		env.run(() => `test-script ${args.file}`)
+		const { default: TestScriptGenerator } = await import('../generator/test-script/index')
```
phileba, line 14: from directory, the index file import as default
```suggestion
		const { default: TestScriptGenerator } = await import('../generator/test-script')
```

packages/cli/src/cmd/generate.ts
```diff
@@ -1,14 +1,19 @@
...
+		env.registerStub(TestScriptGenerator as any, 'element/test-script')
+		env.run(['element/test-script', args.file], () => `test-script ${args.file}`)
```
phileba, line 16: in callback function we can write down some log to let user know the process done
```suggestion
		env.run(['element/test-script', args.file], () => {
		  console.log('Generate success')
		})
```

PR #173, opened 2020-06-03
feat: enhance logic of element options #172

### LINKED ISSUE #172: Enhance logic of element options
Enhance logic of function setupDelayOverrides in run.ts file

ai-flood:
resolve #172


packages/cli/src/cmd/run.ts
```diff
@@ -37,17 +37,16 @@ interface RunArguments extends Arguments {
 function setupDelayOverrides(args: RunArguments, testSettingOverrides: TestSettings) {
 	if (testSettingOverrides == null) testSettingOverrides = {}
 
+	testSettingOverrides.actionDelay = args.actionDelay ?? 0
+	testSettingOverrides.stepDelay = args.stepDelay ?? 0
+
 	if (args.fastForward ?? false) {
 		testSettingOverrides.stepDelay = args.fastForward
 		testSettingOverrides.actionDelay = args.fastForward
```
ivanvanderbyl, line 45: Shouldn't these have a numeric value? What type/value is args.fastForward?

phileba: yeah, I think its type is `boolean` only, the same as `args.slowMo`. So it should be
```
if (args.fastForward) {
	testSettingOverrides.stepDelay = 1
	testSettingOverrides.actionDelay = 1
} else if (args.slowMo) {
	testSettingOverrides.stepDelay = 10
	testSettingOverrides.actionDelay = 10
}
```
we also can custom the `stepDelay` and `actionDelay` by adding the value for them.

packages/cli/src/cmd/run.ts
```diff
@@ -37,17 +37,16 @@ interface RunArguments extends Arguments {
...
-		testSettingOverrides.actionDelay = args.fastForward
-	} else if (args.slowMo ?? false) {
-		testSettingOverrides.stepDelay = args.slowMo ?? testSettingOverrides.stepDelay
-		testSettingOverrides.actionDelay = args.slowMo ?? testSettingOverrides.actionDelay
-	}
-
-	testSettingOverrides.actionDelay = args.actionDelay ?? testSettingOverrides.actionDelay
-	testSettingOverrides.actionDelay = args.stepDelay ?? testSettingOverrides.stepDelay
+	testSettingOverrides.actionDelay = args.actionDelay ?? 0
+	testSettingOverrides.stepDelay = args.stepDelay ?? 0
```
phileba, line 41: I think we should handle the negative case, let say if the user input value -1?
```suggestion
	testSettingOverrides.actionDelay = args.actionDelay && args.actionDelay > 0 ? args.actionDelay : 0
	testSettingOverrides.stepDelay = args.stepDelay && args.stepDelay > 0 ? args.stepDelay : 0
```

PR #174, opened 2020-06-03
feature/add step repeatable

### LINKED ISSUE #68: Ability to mark a step as repeatable

phileba:
### What does it do?
Allow user repeat step by number or a condition function

```ts
step.repeat(5, "Increment counter", async browser => { ... })
step.while(() => {return true}, "Increment counter", async browser => { ... })
```
resolve #68

### DIFF
--- a/packages/core/src/runtime/Settings.ts
+++ b/packages/core/src/runtime/Settings.ts
@@ -234,6 +234,11 @@ export interface TestSettings {
 	 * The list of Chromium flags can be found at https://peter.sh/experiments/chromium-command-line-switches/
 	 */
 	launchArgs?: string[]
+
+	/**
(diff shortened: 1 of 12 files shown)


packages/core/src/runtime/Settings.ts
```diff
@@ -234,6 +234,11 @@ export interface TestSettings {
 	 * The list of Chromium flags can be found at https://peter.sh/experiments/chromium-command-line-switches/
 	 */
 	launchArgs?: string[]
+
+	/**
+	 * Define the loop count of the recovery step
+	 */
+	maxRecovery?: number
```
ivanvanderbyl, line 241: This is a bit ambiguous, perhaps name it: `recoveryTries`

PR #175, opened 2020-06-03
feature/add step repeatable

### LINKED ISSUE #68: Ability to mark a step as repeatable

phileba:
### What does it do?
Allow user repeat step by number or a condition function

```ts
step.repeat(5, "Increment counter", async browser => { ... })
step.while(() => {return true}, "Increment counter", async browser => { ... })
```
resolve #68

### DIFF
--- a/examples/condition/test.ts
+++ b/examples/condition/test.ts
@@ -41,7 +41,7 @@ export default () => {
 	})
 
 	step.recovery('Step 3', async browser => {
-		console.log('Recovery step 3 done, throw an error')
+		console.log('Recovery step 3 done, continue the next step')
 		return RecoverWith.CONTINUE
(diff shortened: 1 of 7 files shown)


ivanvanderbyl:
This looks pretty good. Can we make it so that while can accept a promise which resolves to true? And a browser argument, so you can have a step which checks something on the page?

phileba:
> This looks pretty good. Can we make it so that while we can accept a promise which resolves to true? And a browser argument, so you can have a step that checks something on the page?

yep, it already supported, you can see this one:
```
export type ConditionFn = (this: void, browser: Browser) => boolean | Promise<boolean>
stepWhile?: {
	predicate: ConditionFn
}
```

PR #177, opened 2020-06-04
[FLO-1690] CLI actions for launching tests on Flood

### LINKED ISSUE #142: CLI actions for launching tests on Flood
It'd be neat to be able to launch a Flood from the Element CLI. Something like:

```shell
element flood run script.ts
```

We'd need some other details for handling login token and maybe listing running floods and results.

HieuLeKMS:
**Authenticate:** `element flood authenticate <flood_api_token>`

**List projects:** `element flood project ls`

**Select a project to use:** `element flood use "<project_name>"`

**Show current project:** `element flood project`

**Run a flood:** `element flood run <script> [--options]`

**NOTE:**
- Added 2 new packages `node-fetch` & `form-data` to send test scripts to Flood using Flood's API v2
- Can run flood using token & project (name or id) from user environment
e.g. `FLOOD_API_TOKEN=<token> PROJECT=<project_id> element flood run ...`
- There will be another PR to make it possible to launch floods using config file, once https://github.com/flood-io/element/pull/180 was merged into `beta`

resolve #142


packages/cli/src/generator/flood/grids-select/index.ts
```diff
@@ -0,0 +1,96 @@
...
+			options: { virtualUser },
+			answers: { grids },
+		} = this
+		const totalUsers = virtualUser * grids.length
+
+		return `You're about to simulate ${totalUsers} total user${
+			totalUsers === 1 ? '' : 's'
+		} across ${grids.length} grid${grids.length === 1 ? '' : 's'}. Estimated usage will be ${
+			this._vuh
+		} VUH. Are you sure?`
```
phileba, line 44: it's better to use pluralize, e.g: https://www.npmjs.com/package/pluralize

HieuLeKMS: let's not add not-really-necessary external packages, when they won't be used much, or things can be easily done without them

phileba: cool

packages/cli/src/cmd/flood/run.ts
```diff
@@ -0,0 +1,76 @@
...
+		const { default: InfrastructureSelect } = args.hosted
+			? await import('../../generator/flood/grids-select')
+			: await import('../../generator/flood/regions-select')
+		const env = YoEnv.createEnv()
+
+		env.on('error', err => {
+			console.error(chalk.redBright(err))
+			process.exit(1)
+		})
+		env.registerStub(InfrastructureSelect as any, 'element/infrastructure-select')
```
ivanvanderbyl, line 30: Isn't there a type for this?

packages/cli/src/cmd/flood/run.ts
```diff
@@ -0,0 +1,76 @@
...
+				group: 'Flood options:',
+				describe: 'Run a flood using hosted grid, otherwise using on-demand grid',
+				type: 'boolean',
+				default: false,
+			})
+			.option('virtual-user', {
+				group: 'Flood options:',
+				alias: 'vu',
+				describe:
+					'Number of users per region (when using hosted grid) or per node (when using on-demand grid)',
```
ivanvanderbyl, line 45: This description sounds incorrect. Isn't it the other way around?

packages/cli/src/cmd/flood/run.ts
HieuLeKMS: There is. `registerStub` receives a class that extends `Generator` of yeoman. But when type-checking, instead of a class, it requires an instance of that class. So I converted it to any to bypass type-checking

packages/cli/src/cmd/flood/run.ts
HieuLeKMS: You're right. I fixed it

PR #184, opened 2020-06-13
feat: Support for Microsoft Playwright

phileba:
Instead of running the test on Chrome only, now we should support users running a test on other kinds of browsers like Firefox, Safari. So Playwright is a good choice for replacing.
https://github.com/microsoft/playwright


packages/core/package.json
```diff
@@ -48,6 +48,7 @@
 		"ksuid": "^1.2.0",
 		"micromatch": "^4.0.2",
 		"puppeteer": "^2.1.1",
```
ai-flood, line 50: Why haven't we removed puppeteer yet?

packages/core/src/page/types.ts
```diff
@@ -19,16 +13,23 @@ export interface LocatorBuilder {
 	toString(): string
 }
 
+export type EvaluateFn<T = any> = string | ((arg1: T, ...args: any[]) => any)
+export interface PageGoToOptions {
```
HieuLeKMS, line 17: I think `NavigationOptions` sounds better, and that's what we're currently using, with puppeteer

examples/internet-herokuapp/test.ts
```diff
@@ -0,0 +1 @@
+class Test {}
```
HieuLeKMS, line 1: what's this for?

packages/core/src/runtime/Settings.ts
```diff
@@ -48,7 +47,7 @@ export type ConsoleMethod = 'log' | 'info' | 'debug' | 'warn' | 'error'
  * puppeteer | (Default) The browser bundled with [puppeteer]. It is a curated version of [chromium](https://www.chromium.org) (the open source version of Google Chrome). Using the puppeteer-bundled Chromium ensures the best compatibility with puppeteer, but lacks some features such as video support.
  * stable | The latest version of [Google Chrome](https://www.chromium.org/). Google Chrome has more features than chromium, but isn't tested as thoroughly against puppeteer, which can result in intermittent errors. If you don't need the extra features, please use `bundled`.
  */
-export type ChromeVersion = 'puppeteer' | 'stable'
+export type BrowserType = BROWSER_TYPE
```
HieuLeKMS, line 50: The description above should be updated

HieuLeKMS: Also I don't see BrowserType is used anywhere

phileba: you can use `BrowserType` in TestSetting, to define which browser will be run for the test

examples/internet-herokuapp/test.ts
phileba: it's just a simple test file for the `18-File_Upload.ts` => `Test: 03 - Choose File`

PR #186, opened 2020-06-17
Features/support functions with hook #179

### LINKED ISSUE #179: Support some functions with hooks
Should provide an ability to run some functions before all step, before each step, after all step and after each step, e.g:
```
afterAll(fn, timeout)
afterEach(fn, timeout)
beforeAll(fn, timeout)
beforeEach(fn, timeout)
```

ai-flood:
resolve #179

### DIFF
--- a/examples/hook/test.ts
+++ b/examples/hook/test.ts
@@ -2,13 +2,15 @@ import { TestSettings, step, afterAll, afterEach, beforeAll, beforeEach } from '
 
 export const settings: TestSettings = {
 	loopCount: 1,
+	waitTimeout: 30,
 }
 export default () => {
-	beforeAll(() => {
+	beforeAll(async browser => {
+		await browser.visit('https://challenge.flood.io')
 		console.log('before all is running ....')
 	}, 1)
 
-	beforeEach(async () => {
+	beforeEach(async browser => {
 		console.log('before each is running ....')
 	}, 1)
 
@@ -19,11 +21,11 @@ export default () => {
 		console.log('Step 1')
 	})
 
-	afterEach(async () => {
+	afterEach(async browser => {
 		console.log('after each is running ....')
 	}, 1)
 
-	afterAll(async () => {
+	afterAll(async browser => {
 		console.log('after all is running ....')
 	})
 }
--- a/examples/hook/test.ts
+++ b/examples/hook/test.ts
@@ -1,0 +1,13 @@
+import { TestSettings, step } from '@flood/element'
+
+export const settings: TestSettings = {
+	loopCount: 1,
+}
+export default () => {
+	step('Step 1', async browser => {
+		console.log('Step 1')
+	})
+	step('Step 2', async browser => {
+		console.log('Step 2')
+	})
+}
--- a/examples/hook/test.ts
+++ b/examples/hook/test.ts
@@ -6,11 +6,11 @@ export const settings: TestSettings = {
 }
 export default () => {
 	beforeAll(async browser => {
-		await browser.visit('https://challenge.flood.io')
+		//await browser.visit('https://challenge.flood.io')
 		console.log('before all is running ....')
 	}, 1)
 
-	beforeEach(async browser => {
+	beforeEach(async () => {
 		console.log('before each is running ....')
 	}, 1)
 
@@ -21,11 +21,11 @@ export default () => {
 		console.log('Step 1')
 	})
 
-	afterEach(async browser => {
+	afterEach(async () => {
 		console.log('after each is running ....')
 	}, 1)
 
-	afterAll(async browser => {
+	afterAll(async () => {
 		console.log('after all is running ....')
 	})
 }
--- a/examples/hook/test.ts
+++ b/examples/hook/test.ts
@@ -1,13 +1,41 @@
-import { TestSettings, step } from '@flood/element'
+import { TestSettings, step, afterAll, afterEach, beforeAll, beforeEach } from '@flood/element'
 
 export const settings: TestSettings = {
 	loopCount: 1,
 }
 export default () => {
+	beforeAll(() => {
+		console.log('before all 0 is running ....')
+	}, 300)
+	beforeAll(() => {
+		console.log('before all 1 is running ....')
+	}, 300)
+
+	beforeEach(async () => {
+		console.log('before each 0 is running ....')
+	}, 300)
+	beforeEach(async () => {
+		console.log('before each 1 is running ....')
+	}, 300)
+
+	step('Step 0', async browser => {
+		console.log('Step 0')
+	})
 	step('Step 1', async browser => {
 		console.log('Step 1')
 	})
-	step('Step 2', async browser => {
-		console.log('Step 2')
-	})
+
+	afterAll(async () => {
+		console.log('after all 0 is running ....')
+	}, 300)
+	afterAll(async () => {
+		console.log('after all 1 is running ....')
+	}, 300)
+
+	afterEach(async () => {
+		console.log('after each 0 is running ....')
+	}, 300)
+	afterEach(async () => {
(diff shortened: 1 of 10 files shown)


packages/core/src/runtime/Hook.ts
```diff
@@ -0,0 +1,31 @@
...
+	afterEach: HookBase[]
+	beforeAll: HookBase[]
+	beforeEach: HookBase[]
+}
+
+/**
+ * @internal
+ */
+export function normalizeHookBase(hookBase: HookBase): HookBase {
+	// Convert user inputted seconds to milliseconds
```
HieuLeKMS, line 23: This looks more like convert ms to s

packages/core/src/runtime/Hook.ts
```diff
@@ -0,0 +1,31 @@
...
+	if (typeof hookBase.timeout === 'number' && hookBase.timeout > 1e3) {
```
HieuLeKMS, line 24: What if `hookBase.timeout === 1e3`?

ai-flood: Just keep logic related to timeout same as the current logic of waitTimeout (function normalizeStepOptions of Step.ts)

packages/core/src/runtime/Hook.ts
ai-flood: The reason looks like the under the comment

packages/core/src/runtime/EvaluatedScript.ts
```diff
@@ -284,6 +316,12 @@ export class EvaluatedScript implements TestScriptErrorMapper, EvaluatedScriptLi
 			// Supports either 2 or 3 args
 			step,
 
+			//Hook
+			afterAll: captureHookWithAfterAll,
+			afterEach: captureHookWithAfterEach,
+			beforeAll: captureHookWithBeforeAll,
+			beforeEach: captureHookWithBeforeEach,
+
 			// Actual implementation of @flood/chrome
```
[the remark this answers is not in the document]
ai-flood: The comment doesn't relate to current change so I think we should it is not the problem

packages/core/src/runtime/EvaluatedScript.ts
```diff
@@ -284,6 +316,12 @@ export class EvaluatedScript implements TestScriptErrorMapper, EvaluatedScriptLi
 			// Supports either 2 or 3 args
 			step,
 
+			//Hook
+			afterAll: captureHookWithAfterAll,
+			afterEach: captureHookWithAfterEach,
+			beforeAll: captureHookWithBeforeAll,
+			beforeEach: captureHookWithBeforeEach,
```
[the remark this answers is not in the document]
ai-flood: When we replace captureHookWithAfterAll with afterAll, the function captureHookWithAfterAll also renames afterAll. I just want to the function name more meaningful. And if the reason why I give that can't satisfy you, I will resolve it based on your idea

packages/core/src/runtime/Hook.ts
```diff
@@ -0,0 +1,31 @@
...
+export declare function beforeAll(fnc: HookFn, timeout?: number): void
+export declare function beforeEach(fnc: HookFn, timeout?: number): void
+
+export type HookFn = (this: void) => any
+export interface HookBase {
+	fnc: HookFn
+	timeout?: number
+}
+
+export type Hook = {
```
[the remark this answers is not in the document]
ai-flood: Based on the requirement, a test script can have more than 1 afterAll function, my idea is we need an object included property names the same as the function names I defined before and the type of the property is a type of list.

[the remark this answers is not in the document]
HieuLeKMS: I mean we can make `Hook` an interface as well, just to be consistent

packages/core/src/runtime/EvaluatedScript.ts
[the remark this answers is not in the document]
phileba: yeah I think we should follow as @HieuLeKMS suggested, `afterAll` name itself has meaning and we can understand what it does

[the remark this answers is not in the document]
ai-flood: Renamed function names

packages/core/src/runtime/Hook.ts
```diff
@@ -0,0 +1,31 @@
...
+
+/**
+ * @internal
+ */
+export function normalizeHookBase(hookBase: HookBase): HookBase {
+	// Convert user inputted seconds to milliseconds
+	if (typeof hookBase.timeout === 'number' && hookBase.timeout > 1e3) {
+		hookBase.timeout = hookBase.timeout / 1e3
+	} else if (Number(hookBase.timeout) === 0 || !hookBase.timeout) {
+		hookBase.timeout = 30
```
phileba, line 27: I think right now it should be
```suggestion
      const { timeout } = hookBase
	hookBase.timeout = timeout &&  timeout > 0 ? timeout : 30 * 1e3
```
in the next enhancement ticket, we can remove this function and the timeout value should evaluate by something likes
```
function captureHookWithAfterAll(...args: any[]) {
	const [fnc, timeout] = args
	hook.afterAll.push({ fnc, ms(timeout)})
}
```

packages/core/src/runtime/Hook.ts
```diff
@@ -0,0 +1,31 @@
+export declare function afterAll(fnc: HookFn, timeout?: number): void
+export declare function afterEach(fnc: HookFn, timeout?: number): void
+export declare function beforeAll(fnc: HookFn, timeout?: number): void
+export declare function beforeEach(fnc: HookFn, timeout?: number): void
+
+export type HookFn = (this: void) => any
```
phileba, line 6: I think we should allow passing `browser` variable in case if the user wants to do something/checking element
```suggestion
export type HookFn = (this: void, browser: Browser) => Promise<any>
```

packages/core/src/runtime/Test.ts
```diff
@@ -160,6 +163,9 @@ export default class Test implements ITest {
 			debug('running this.before(browser)')
 			await testObserver.before(this)
 
+			debug('running hook function: beforeAll')
+			this.hook.beforeAll.forEach(item => this.runHookFunction(item))
```
phileba, line 167: `forEach` is synchronous so it will not wait for the function `runHookFunction` runs completely so the test will fail. In this case, I think we should write a common function like `runHookFnc`
```suggestion
	const runHookFnc = async (hooks: HookBase[]) => {
	    try {
	        for (const hook of hooks){
	           await hook.fnc.call(null, browser)
	        }
	    } catch (err){
	        throw err
	    }
	}
	...
	//calling
	await runHookFnc(this.hook.beforeAll)
```

packages/core/src/runtime/Test.ts
```diff
@@ -396,4 +411,9 @@ export default class Test implements ITest {
 	newTrace(step: Step): ObjectTrace {
 		return new ObjectTrace(this.script.runEnv.workRoot, step.name)
 	}
+
+	private runHookFunction(hookBase: HookBase): void {
+		hookBase.fnc()
+		setTimeout(this.runHookFunction, hookBase.timeout)
```
phileba, line 417: I'm not sure why we need this?

examples/hook/test.ts
```diff
@@ -0,0 +1,29 @@
+import { TestSettings, step, afterAll, afterEach, beforeAll, beforeEach } from '@flood/element'
+
+export const settings: TestSettings = {
+	loopCount: 1,
+}
+export default () => {
+	beforeAll(() => {
```
phileba, line 7: how is it going?
```
beforeAll(async browser => {
  await browser.wait(Until.elementIsVisible(locator))
}, 1000)
```

PR #188, opened 2020-06-19
Frame not found if its name is empty & is found by ElementHandle

### LINKED ISSUE #182: Error thrown when switching to iframe
**Describe the bug**
When attempting to switch context to an iframe that contains some form fields i'm recieving the folllowing error

```
+66s info: ===> Step 'Test: User enters credit card information'
+66s info: ---> wait()
+66s info: ---> wait()
step error -> failed
+66s error:


Error: Could not match frame by name or id: 'easyXDM_default4069_provider'


Detail:
Documentation needed! Please report this at https://github.com/flood-io/element/issues so that we can improve Element!

+66s error: xxxx Step 'Test: User enters credit card information' failed
+66s info: ---> Step 'Test: User enters credit card information' finished
failed, bailing out of steps
+66s info: Iteration completed in 64858ms (walltime)
+66s info: Test completed after 1 iterations
```

**To Reproduce**
1. Go to 'https://www.blinds.com' and simply search for a product add to cart and get all the way to credit card checkout screen where you enter in your information

**Expected behavior**
I'm expected to be able to enter credit card details into the iframe presented in the page

**Desktop (please complete the following information):**
 - OS: macOS
 - Node.js version: 14.4
 - Version: 1.2.3

**code**
```
		await browser.wait(Until.urlContains('Ordering/Checkout/Billing'))
		let iframe = await browser.findElement(By.tagName('iframe'))
		await browser.switchTo().frame(iframe)
```

HieuLeKMS:
Currently, a frame cannot be found if its name is empty and we find it by `ElementHandle`. In that case, we find it from `window.frames`

resolve #182

### DIFF
--- a/packages/core/src/page/TargetLocator.ts
+++ b/packages/core/src/page/TargetLocator.ts
@@ -25,6 +25,19 @@ export class TargetLocator implements ITargetLocator {
 		return new ElementHandle(element).initErrorString()
 	}
 
+	private async findFrameFromWindow(id: string | null): Promise<Frame | undefined> {
+		const frames = getFrames(this.currentPage.frames())
+		const frameElementName = await this.currentPage.evaluate((id: string) => {
+			// NOTE typescript lib.dom lacks proper index signature for frames: Window to work
+			const frame = Array.from(window.frames).find(frame => frame.frameElement.id === id)
+
+			if (!frame) throw Error(`No frame found with id=${id}`)
+			return frame.name
+		}, id)
+
+		return frames.find(frame => frame.name() === frameElementName)
+	}
+
 	/**
 	 * Navigates to the topmost frame
 	 */
@@ -49,7 +62,7 @@ export class TargetLocator implements ITargetLocator {
 			return
 		}
 
-		let nextFrame: Frame | null
+		let nextFrame: Frame | undefined
 
 		const frames = getFrames(this.currentPage.frames())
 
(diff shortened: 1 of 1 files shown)


packages/core/src/page/TargetLocator.ts
```diff
@@ -63,24 +76,15 @@ export class TargetLocator implements ITargetLocator {
 				return frame.name || frame.id
```
phileba, line 76: if this one returns `frame.id` so this check on line 79
```
nextFrame = frames.find(frame => frame.name() === frameElementName)
```
is still right?

packages/core/src/page/TargetLocator.ts
```diff
@@ -96,7 +100,10 @@ export class TargetLocator implements ITargetLocator {
 			let name = await id.getProperty('name')
```
phileba, line 100:
```suggestion
			let nameOrId = await id.getProperty('name')
```

packages/core/src/page/TargetLocator.ts
```diff
@@ -63,24 +76,15 @@ export class TargetLocator implements ITargetLocator {
 				return frame.name || frame.id
 			}, id)
 
-			nextFrame = frames.find(frame => frame.name() === frameElementName) || null
+			nextFrame = frames.find(frame => frame.name() === frameElementName)
 			if (!nextFrame) throw new Error(`Could not match frame by name or id: '${frameElementName}'`)
 
 			this.applyFrame(nextFrame)
 		} else if (typeof id === 'string') {
```
phileba, line 83: I think we should change the `id` to another name that is more general

packages/core/src/page/TargetLocator.ts
HieuLeKMS: it should be right
if the frame's name is empty, `frame.name()` will return the frame's id instead
https://pptr.dev/#?product=Puppeteer&version=v4.0.0&show=api-framename

packages/core/src/page/TargetLocator.ts
HieuLeKMS: it's the value of `getProperty('name')` so just `name` is fine

packages/core/src/page/TargetLocator.ts
HieuLeKMS: it's part of the API so I don't think I should change it

packages/core/src/page/TargetLocator.ts
phileba: I see we get the value of id in the next line

HieuLeKMS: I still think it's fine since it's compared with `frame.name()`

PR #198, opened 2020-06-29
Remove suite

HieuLeKMS:
`suite` was an old idea to support test data, now deprecated.
This PR is for removing `suite` & updating examples which use `suite`


phileba:
@HieuLeKMS can you remove this one 
```
const context = {
        ...
	suite: captureSuite,
	...
}
```
in `EvaluatedScript.ts - evaluate()` then it will be good

PR #199, opened 2020-06-29
Features/global recovery step #181

### LINKED ISSUE #181: Should provide a global recovery step
We should have a global recovery step to provide an option when step failed/error

```
step.recovery(async browser => {
// do stuff
})
```

ai-flood:
resolve #181

### DIFF
--- a/examples/condition/test.ts
+++ b/examples/condition/test.ts
@@ -86,7 +86,6 @@ export default () => {
 
 	let count = 0
 	step('Step 1', async browser => {
-		console.log('Throw error step 1')
 		throw Error('Throw error step 1')
 	})
 	// step.recovery(async () => {
@@ -97,7 +96,11 @@ export default () => {
 	// 	console.log('Done step 1')
 	// 	return RecoverWith.RETRY
 	// })
-	step.recovery('Step 1', { recoveryTries: 3 }, async browser => {
+	// step.recovery('Step 1', { recoveryTries: 2 }, async browser => {
+	// 	return RecoverWith.RETRY
+	// })
+
+	step.globalRecovery({ recoveryTries: 2 }, async browser => {
 		return RecoverWith.RETRY
 	})
 }
--- a/examples/condition/test.ts
+++ b/examples/condition/test.ts
@@ -5,102 +5,93 @@ export const settings: TestSettings = {
 }
 
 export default () => {
-	// step.once('Test once', async browser => {
-	// 	console.log('Step once runs 1 time')
-	// })
-	// step.if(
-	// 	() => true,
-	// 	'Test if',
-	// 	async browser => {
-	// 		console.log('Step if runs with condition true')
-	// 	},
-	// )
-	// step.unless(
-	// 	() => false,
-	// 	'Test unless',
-	// 	async browser => {
-	// 		console.log('Test unless runs with condition false')
-	// 	},
-	// )
-	// step.skip('Test skip', async browser => {
-	// 	console.log('This step will not be run')
-	// })
-	// step('Step pending')
+	step.once('Test once', async browser => {
+		console.log('Step once runs 1 time')
+	})
+	step.if(
+		() => true,
+		'Test if',
+		async browser => {
+			console.log('Step if runs with condition true')
+		},
+	)
+	step.unless(
+		() => false,
+		'Test unless',
+		async browser => {
+			console.log('Test unless runs with condition false')
+		},
+	)
+	step.skip('Test skip', async browser => {
+		console.log('This step will not be run')
+	})
 
-	// let retry = false
-	// let restart = false
+	let retry = false
+	let restart = false
 
-	// step.recovery('Step 1', async browser => {
-	// 	console.log('Recovery step 1 done, retry step 1')
-	// 	return RecoverWith.RETRY
-	// })
+	step.recovery('Step 1', async browser => {
+		console.log('Recovery step 1 done, retry step 1')
+		return RecoverWith.RETRY
+	})
 
-	// step.recovery('Step 2', async browser => {
-	// 	console.log('Recovery step 2 done, restart step 1')
-	// 	return RecoverWith.RESTART
-	// })
+	step.recovery('Step 2', async browser => {
+		console.log('Recovery step 2 done, restart step 1')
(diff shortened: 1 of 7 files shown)


packages/core/src/runtime/EvaluatedScript.ts
```diff
@@ -68,6 +69,7 @@ function createVirtualMachine(floodElementActual: any, root?: string): NodeVM {
 export class EvaluatedScript implements TestScriptErrorMapper, EvaluatedScriptLike {
 	public steps: Step[]
 	public recoverySteps: StepRecoveryObject
+	public globalRecoveryStep: GlobalRecoveryObject | null
```
phileba, line 72: can we store the global recovery step object in the `recoverySteps`?

packages/core/src/runtime/EvaluatedScript.ts
```diff
@@ -202,23 +205,23 @@ export class EvaluatedScript implements TestScriptErrorMapper, EvaluatedScriptLi
 			},
 		)
 
-		const step: StepExtended = (name: string, ...optionsOrFn: any[]) => {
-			const [option, fn] = extractOptionsAndCallback(optionsOrFn)
+		const step: StepExtended = (...nameOrOptionsOrFn: any[]) => {
+			const [name, option, fn] = extractNameAndOptionsAndCallback(nameOrOptionsOrFn)
```
phileba, line 209: `extractNameAndOptionsAndCallback` seems too long, I think we can call `extractStep`?

packages/core/src/runtime/Step.ts
```diff
@@ -115,14 +122,25 @@ export type StepOptions = {
...
 	} else if (args.length === 2) {
-		const [options, fn] = args as [StepOptions, TestFn]
+		const [name, fn] = args as [string, TestFn]
+		if (typeof name === 'string') {
+			return [name, {}, fn]
+		} else {
+			const [options, fn] = args as [StepOptions, TestFn]
+			const { waitTimeout, waitUntil, recoveryTries } = options
+			return ['global', { waitTimeout, waitUntil, recoveryTries }, fn]
+		}
```
phileba, line 139:
```suggestion
		if (typeof name === 'string') {
			return [name, {}, fn]
		} 
		const [options, fn] = args as [StepOptions, TestFn]
		const { waitTimeout, waitUntil, recoveryTries } = options
		return ['global', { waitTimeout, waitUntil, recoveryTries }, fn]
		
```

packages/core/src/runtime/Test.ts
```diff
@@ -149,8 +151,15 @@ export default class Test implements ITest {
...
-		if (!stepRecover) return false
+		let stepRecover = this.recoverySteps[step.name]
+		if (!stepRecover) {
+			if (this.globalRecoveryStep) {
+				stepRecover = this.globalRecoveryStep
+				if (!stepRecover) return false
+			} else {
+				return false
+			}
+		}
```
phileba, line 162:
```suggestion
		if (!stepRecover && this.globalRecoveryStep) {
			stepRecover = this.globalRecoveryStep	
		} else {
			return false
		}
```

packages/core/src/runtime/Test.ts
```diff
@@ -149,8 +151,15 @@ export default class Test implements ITest {
 		looper: Looper,
 		browser: BrowserInterface,
 	): Promise<boolean> {
-		const stepRecover = this.recoverySteps[step.name]
-		if (!stepRecover) return false
+		let stepRecover = this.recoverySteps[step.name]
+		if (!stepRecover) {
+			if (this.globalRecoveryStep) {
+				stepRecover = this.globalRecoveryStep
+				if (!stepRecover) return false
```
phileba, line 158: we have checked `if (this.globalRecoveryStep)` so I think we don't need it

packages/core/src/runtime/Step.ts
```diff
@@ -58,6 +59,12 @@ export interface StepRepeatablebase {
 	(count: number, ...optionsOrFn: any[])
 }
 
+export interface StepGlobal {
```
phileba, line 62: in the future, I think we don't have any kind of global step so I think we should specify this name `StepGlobalRecovery`

packages/core/src/runtime/Step.ts
```diff
@@ -58,6 +59,12 @@ export interface StepRepeatablebase {
...
+	(options: StepOptions, testFn: TestFn): void
+	(testFn: TestFn): void
+	(...optionsOrFn: any[]): void
```
phileba, line 65: I think don't need it

packages/core/src/runtime/Step.ts
```diff
@@ -115,14 +122,22 @@ export type StepOptions = {
 	}
 }
 
-export function extractOptionsAndCallback(args: any[]): [Partial<StepOptions>, TestFn] {
-	if (args.length === 0) return [{ pending: true }, () => Promise.resolve()]
+export function extractStep(args: any[]): [string, Partial<StepOptions>, TestFn] {
+	if (args.length === 0) throw new Error('Step called with at least one argument') // [{ pending: true }, () => Promise.resolve()]
 	if (args.length === 1) {
-		return [{}, args[0] as TestFn]
+		return ['global', {}, args[0] as TestFn]
```
ivanvanderbyl, line 128: What is "global" here?

examples/condition/test.ts
```diff
@@ -83,4 +82,12 @@ export default () => {
 			console.log('Step 5')
 		},
 	)
+
+	step('Step 6', async browser => {
+		throw Error('Throw error step 6')
+	})
+	step.recovery({ recoveryTries: 2 }, async browser => {
```
ivanvanderbyl, line 89: This doesn't read well, let's make it `tries`

packages/core/src/runtime/Step.ts
ai-flood: It is a global recovery step that would be returned after extracting arguments of step(To distinguish a global and a local recovery step).
In the normal case, a step is defined: step('StepName', async browser => {}) 
In case we define a recovery step,
- a local recovery step is defined: step.recovery('StepName', async browser => {})
- a global recovery step is defined: step.recovery(async browser => {})

PR #202, opened 2020-07-04
Refactor/breaking dependency cycles

### LINKED ISSUE #176: Fix 'Dependency cycle...' warning on importing file
**Describe the bug**
There are many files that we using dependency cycle import. This may cause an error in the future and I think we need to fix it.

HieuLeKMS:
Breaking dependency cycles as shown below

image

resolve #176


HieuLeKMS:
> This is a huge improvement, great work!!
> 
> It looks like we still have 2 cyclic deps:
> 
> ```shell
> ❯ npx madge --circular --extensions ts ./packages
> npx: installed 150 in 20.125s
> Processed 204 files (1.9s) (32 warnings)
> 
> ✖ Found 2 circular dependencies!
> 
> 1) core/src/driver/puppeteer.ts
> 2) core/src/page/Locator.ts > core/src/page/ElementHandle.ts
> ```

I've just taken a look and see that:
1. There are no cycles involving `core/src/driver/Puppeteer.ts`
2. The cycle `core/src/page/Locator.ts > core/src/page/ElementHandle.ts` has already been fixed, using lazy importimport

I checked using eslint and there are no cycles left, don't know how it's still showing here.

PR #215, opened 2020-07-20
feat: change terminal output style #204

### LINKED ISSUE #204: New terminal output style while running a local test

ai-flood:
resolve #204


packages/core/package.json
```diff
@@ -52,6 +52,7 @@
 		"sanitize-filename": "^1.6.1",
 		"source-map": "^0.7.0",
 		"strftime": "^0.10.0",
+		"table": "^5.4.6",
```
HieuLeKMS, line 55: Since you added a new package, you should also commit `yarn.lock`. Otherwise when publishing to NPM, we'll get an error as in the image 
image

packages/cli/src/cmd/flood/run.ts
```diff
@@ -24,7 +24,7 @@ const cmd: CommandModule = {
 		const env = YoEnv.createEnv()
 
 		env.on('error', err => {
-			console.error(chalk.redBright(err))
```
HieuLeKMS, line 27: why are all these console.error replaced by console.debug?

ai-flood: Because currently as agreed with the BA, a @phileba, only log out the errors of the test script, unnecessary errors will show up in the form of debugging.
And when we implement our own set of loggers, we will enhance them.
In this case, maybe the above error is not related to the test script, so I will revert this change

packages/core/package.json
ai-flood: Updated

packages/core/src/runtime/StepIterator.ts
```diff
@@ -0,0 +1,156 @@
...
+
+	goPreviousStep(): boolean {
+		this.stepCount -= 1
+		return true
+	}
+
+	stepEnd() {
+		this.stepCount = this.steps.length
+		return true
+	}
```
HieuLeKMS, line 31: why these always return true? should these's return type be void instead?
```suggestion
	goNextStep(): void {
		this.stepCount += 1
	}

	goPreviousStep():void {
		this.stepCount -= 1
	}

	stepEnd():void {
		this.stepCount = this.steps.length
	}
```

packages/core/src/runtime/StepIterator.ts
```diff
@@ -0,0 +1,156 @@
...
+			step.prop.recoveryTries = stepRecover.iteration
+		} else {
+			step.prop = { recoveryTries: stepRecover.iteration }
+		}
+		try {
+			const result = await recoveryStep.fn.call(null, browser)
+			const { repeat } = step.options
+			if (result === RecoverWith.CONTINUE) {
+				step.prop.recoveryTries = 0
+				//if (!repeat) return this.goNextStep()
```
HieuLeKMS, line 140: 🔥

packages/core/src/utils/DocumentedError.ts
```diff
@@ -41,16 +41,17 @@ export class DocumentedError extends Error {
 			;(err as DocumentedError).callContext = callCtx
 			return err
 		} else {
-			return DocumentedError.wrapUnhandledError(err, err.message, callCtx)
+			return err
+			//return DocumentedError.wrapUnhandledError(err, err.message, callCtx)
```
HieuLeKMS, line 45: 🔥

packages/core/src/utils/DocumentedError.ts
```diff
@@ -41,16 +41,17 @@ export class DocumentedError extends Error {
...
-			'Documentation needed! Please report this at https://github.com/flood-io/element/issues so that we can improve Element!',
-			callContext,
-			err,
-		)
+		// return new DocumentedError(
+		// 	message || err.message,
+		// 	'Documentation needed! Please report this at https://github.com/flood-io/element/issues so that we can improve Element!',
+		// 	callContext,
+		// 	err,
+		// )
```
HieuLeKMS, line 55: 🔥

packages/element/index.ts
```diff
@@ -124,4 +124,4 @@ export { suite } from '@flood/element-core'
 /**
  * @docPage helpers
  */
-export { RecoverWith } from '@flood/element-core'
+export { RecoverWith, SummaryStep } from '@flood/element-core'
```
HieuLeKMS, line 127: correct me if I'm wrong, `SummaryStep` is used internally to build test result, so I think we don't need to expose it to users
```suggestion
export { RecoverWith } from '@flood/element-core'
```

packages/core/src/runtime/Test.ts
```diff
@@ -53,7 +55,8 @@ export default class Test implements ITest {
 
 	public failed: boolean
 
-	public stepCount: number
+	public summaryStep: SummaryStep[] = []
+	public subTitle: string = ''
```
HieuLeKMS, line 59: eslint should give you the error `Type string trivially inferred from a string literal, remove type annotation` 
```suggestion
	public subTitle = ''
```

packages/core/index.ts
```diff
@@ -113,7 +113,14 @@ export {
...
  * @docPage DSL
  */
-export { step, TestFn, StepFunction, StepOptions, RecoverWith } from './src/runtime/Step'
+export {
+	step,
+	TestFn,
+	StepFunction,
+	StepOptions,
+	RecoverWith,
+	SummaryStep,
```
HieuLeKMS, line 122: If I'm right that `SummaryStep` should not be exposed to users, then it should not be exported here
```suggestion
```

packages/core/src/runtime/StepIterator.ts
```diff
@@ -0,0 +1,154 @@
...
+			} else if (result === RecoverWith.RESTART) {
+				if (repeat) repeat.iteration = 0
+				looper.restartLoop()
+				step.prop.recoveryTries = 0
+				this.stepEnd()
+			} else if (result === RecoverWith.RETRY) {
+				if (repeat) repeat.iteration -= 1
+				else {
+					this.goPreviousStep()
+				}
```
phileba, line 146:
```suggestion
	else this.goPreviousStep()
```

ai-flood:
Already merged into features/report-package branch

PR #226, opened 2020-08-18
[FLO-1756] Example scripts at the-internet/tables

hong-flood:
This PR includes example scripts below:
```
Until.elementTextDoesNotMatch(selectorOrLocator: NullableLocatable, regex: RegExp)
Until.titleIs(title: string)
Until.urlIs(url: string)

By.js(script: EvaluateFn, ...args: any[])

ElementHandle.findElements(locator: Locator | string)
```

### DIFF
--- a/examples/the-internet-more-examples/tables.ts
+++ b/examples/the-internet-more-examples/tables.ts
@@ -1,0 +1,59 @@
+import { ElementHandle } from 'puppeteer'
+import { step, Until, By, TestSettings } from '@flood/element'
+import assert from 'assert'
+
+export const settings: TestSettings = {
+	clearCache: false,
+	disableCache: false,
+	clearCookies: false,
+	loopCount: 1,
+	duration: 1,
+	actionDelay: 2,
+	stepDelay: 2,
+	waitTimeout: 60,
+	screenshotOnFailure: true,
+}
+
+const URL = 'https://the-internet.herokuapp.com'
+
+export default () => {
+	step('Test: Go to the Data Tables page', async browser => {
+		await browser.visit(`${URL}/tables`)
+		// Until.titleIs example
+		await browser.wait(Until.titleIs('The Internet'))
+	})
+
+	step('Test: findElements example', async browser => {
+		// findElements example
+		const tables = await browser.findElements(By.css('.tablesorter'))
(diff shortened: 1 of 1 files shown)


examples/the-internet-more-examples/tables.ts
```diff
@@ -0,0 +1,86 @@
...
+			const table2Id: string = await tables[1].getProperty('id')
+			assert(table1Id === 'table1', 'The id of table 1 must be correct')
+			assert(table2Id === 'table2', 'The id of table 2 must be correct')
+		},
+	)
+
+	step(
+		'Test: Do some sort action with tablesorter',
+		async (browser: Browser): Promise<void> => {
+			const table: Locator = By.id('table2')
```
HieuLeKMS, line 43: My suggestion is that you should remove type annotation in cases like this - when the variable already know its type from the return type of the method. It's unnecessary and makes the code look more dense

examples/the-internet-more-examples/tables.ts
```diff
@@ -0,0 +1,86 @@
...
+			const cellsOfFirstRow: ElementHandle[] = await rows[0].findElements(By.tagName('td'))
+			const lastName: string = await cellsOfFirstRow[0].text()
+			const firstActionEdit: ElementHandle = await rows[0].findElement(By.partialLinkText('edit'))
+			const firstActionDelete: ElementHandle = await rows[0].findElement(
+				By.partialLinkText('delete'),
+			)
+
+			await firstActionEdit.click()
+			await browser.wait(Until.urlIs(`${URL}/tables#edit`))
+			await browser.wait(2)
```
HieuLeKMS, line 57: Are these `await browser.wait(number-of-second)` only used when you debug the script? If so I think you should remove these.

examples/the-internet-more-examples/tables.ts
hong-flood: Got it.

examples/the-internet-more-examples/tables.ts
hong-flood: Yes, it is. I'll remove it.

HieuLeKMS (approved):

PR #236, opened 2020-08-25
[FLO-1757] Example scripts for context menu, drag and drop

hong-flood:
This PR includes example scripts below
```
By.css(selector: string, debugString?: string)

ElementHandle.location()

Mouse.click(x: number, y: number, options: ClickOptions)
Mouse.down(option?: MousePressOptions)
Mouse.up(option?: MousePressOptions)
Mouse.move(x: number, y: number, options?: {steps: number})
Mouse.drag(from: Point, to: Point)

Browser.wait(timeoutOrCondition: Condition | number)
```

### DIFF
--- a/examples/flood-element-challenge.ts
+++ b/examples/flood-element-challenge.ts
@@ -13,13 +13,18 @@ export const settings: TestSettings = {
 }
 
 const URL = 'https://flood-element-challenge.vercel.app/'
-// const URL = 'http://0.0.0.0:8080/'
 
 const getNumOfPages = async (browser: Browser) => {
 	return await browser.evaluate(() => {
 		const nav = document.querySelector('nav[aria-label="pagination navigation"]')
 		const items = nav.getElementsByTagName('li')
 		return items.length === 2 ? 1 : items.length - 2
+	})
+}
+
+const scrollToTop = async (browser: Browser) => {
+	await browser.evaluate(() => {
+		window.scrollTo({ top: 0 })
 	})
 }
 
@@ -172,7 +177,7 @@ export default () => {
 
 	// Challenge 3
 	step('Test: Challenge 3', async browser => {
-		// await browser.page.browserContext().overridePermissions(URL, ['clipboard-read'])
+		await browser.page.browserContext().overridePermissions(URL, ['clipboard-read'])
 
 		const revealButton = await browser.findElement(By.visibleText('Reveal the deal'))
 		await revealButton.click()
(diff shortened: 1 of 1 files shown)


examples/flood-element-challenge.ts
```diff
@@ -0,0 +1,360 @@
...
+}
+
+const maximizePopup = async (browser: Browser) => {
+	const maximizeBtn = await browser.findElement(By.css('button[data-test-maximize]'))
+	await browser.click(maximizeBtn)
+	await browser.wait(0.5)
+}
+
+export default () => {
+	// Before all step
```
HieuLeKMS, line 108: I don't think we need these comments since the hooks and steps' name tell us the exact same thing

examples/flood-element-challenge.ts
```diff
@@ -0,0 +1,360 @@
...
+				}
+			}),
+		)
+
+		await adjustPriceRange(browser, parseInt(targetMinPrice), parseInt(targetMaxPrice))
+		await maximizePopup(browser)
+	})
+
+	step('Test: Challenge 8', async browser => {
+		const subTotalSelector = 'subtotal-price'
```
HieuLeKMS, line 313:
```suggestion
		const subTotalPrice = await browser.findElement(By.id('subtotal-price'))
```

examples/flood-element-challenge.ts
```diff
@@ -0,0 +1,360 @@
...
+		if (currentMaxPrice > targetMaxPrice) {
+			await browser.page.keyboard.press(Key.ARROW_LEFT)
+			currentMaxPrice--
+		} else {
+			await browser.page.keyboard.press(Key.ARROW_RIGHT)
+			currentMaxPrice++
+		}
+	}
+
+	// await browser.wait(0.5)
```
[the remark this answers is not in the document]
hong-flood: Resolved

examples/flood-element-challenge.ts
hong-flood: Resolved.

examples/flood-element-challenge.ts
hong-flood: I declare this variable because I use the string `subtotal-price` 3 times in this step. Can I keep it?

HieuLeKMS: I saw that you used it to find the element 3 times in the exact same way, so I suggested that you find the element first. If I'm wrong then please just resolve this

examples/flood-element-challenge.ts
```diff
@@ -0,0 +1,356 @@
...
+		const checkBtn = await modal.findElement(By.visibleText('CHECK'))
+		checkBtn?.click()
+		await browser.wait(0.5)
+
+		const nextBtn = await modal.findElement(By.visibleText('NEXT'))
+		nextBtn?.click()
+		await browser.wait(0.5)
+	})
+
+	// Challenge 1
```
HieuLeKMS, line 131: can you remove this comment and the ones below as well? as I mentioned in my previous comment, the step's name is clear enough

examples/flood-element-challenge.ts
hong-flood: Got it. I find the element first then get the text again after adding or removing product.

PR #238, opened 2020-08-26
fix: fix several bugs found on element api

HieuLeKMS:
Fix the following bugs:
- ElementHandle.isSelected() always false
- ElementHandle.isSelectable() always false
- ElementSelectedCondition's pageFunc always returns false
- step that has data in TestFn is not defined
- Mouse.click's overloads behave differently

### DIFF
--- a/extern/dogfood/dockerfiles/watirspec/html/wait.html
+++ b/extern/dogfood/dockerfiles/watirspec/html/wait.html
@@ -1,98 +1,98 @@
 <!DOCTYPE html>
 <html>
-  <head>
-    <title>wait test</title>
-    <script type="text/javascript" charset="utf-8">
-      function setTimeoutAddDisplay(id, current_id, timeout) {
-        setTimeoutAdd(id, 'none', id, current_id, timeout)
-        setTimeoutDisplay(id, 'block', timeout * 2)
-      }
+	<head>
+		<title>wait test</title>
+		<script type="text/javascript" charset="utf-8">
+			function setTimeoutAddDisplay(id, current_id, timeout) {
+				setTimeoutAdd(id, 'none', id, current_id, timeout)
+				setTimeoutDisplay(id, 'block', timeout * 2)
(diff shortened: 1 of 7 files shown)


extern/dogfood/dockerfiles/watirspec/html/wait.html
```diff
@@ -1,98 +1,98 @@
...
+				}, timeout)
+			}
 
-      function setTimeoutAdd(id, display, text, existing_id, timeout) {
-        var newDiv = document.createElement('div')
-        newDiv.setAttribute('id', id)
-        newDiv.style.display = display
-        newDiv.appendChild(document.createTextNode(text))
+			function setTimeoutAdd(id, display, text, existing_id, timeout) {
+				var newDiv = document.createElement('div')
```
phileba, line 18: can we replace all `var` by `let` or `const`?

packages/core/src/page/conditions/ElementSelectedCondition.ts
```diff
@@ -21,19 +21,18 @@ export class ElementSelectedCondition extends ElementCondition {
 			}
 
 			if (tagName === 'INPUT') {
-				const type = tagName.toLowerCase()
-				return type == 'checkbox' || type == 'radio'
+				const type = node.getAttribute('type')
```
phileba, line 24: do we need `await node.getAttribute('type')`?

HieuLeKMS: no need for await since it does not return a promise `Element.getAttribute(qualifiedName: string): string | null`

extern/dogfood/dockerfiles/watirspec/html/wait.html
HieuLeKMS: it's in an html file so let's stick with es5 to be safe

phileba (approved):

PR #253, opened 2020-09-01
Update global.md

carltondickson:
Remove mispelt duplicate


phileba:
@carltondickson can you change the target repo name from `master` to `beta` then we can review this.

phileba:
close due duplicate with in https://github.com/flood-io/element/pull/256

Issue #237, opened 2020-08-26
Setting consoleFilter does not work


sonnguyenmau:

phileba:
fixed on the latest version

PR #254, opened 2020-09-02
fix: apply the setting's consoleFilter setting to Observer

brendanoffer:
Hi 👋 , It appears the `consoleFilter` setting is not being applied to the Observer, although it is being referenced. 

It appears we are not the only ones having this [issue](https://github.com/flood-io/element/issues/237).

### DIFF
--- a/packages/core/src/runtime/Observer.ts
+++ b/packages/core/src/runtime/Observer.ts
@@ -14,13 +14,11 @@ interface RequestEvent extends Event {
 }
 
 export default class Observer {
-	public consoleFilters: ConsoleMethod[] = []
-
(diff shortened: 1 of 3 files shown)


phileba:
@brendanoffer can you change the target branch to `flood-io:beta` so we can take a review.

phileba:
> Hi 👋 , It appears the `consoleFilter` setting is not being applied to the Observer, although it is being referenced.
> 
> It appears we are not the only ones having this [issue](https://github.com/flood-io/element/issues/237).

we found (https://github.com/flood-io/element/issues/237) and fixed already.
Thanks for your contribution.

PR #263, opened 2020-09-04
fix: error publishing Element to Homebrew

HieuLeKMS:
Using https://github.com/marketplace/actions/homebrew-tap in both stable and beta workflows to publish the release to Homebrew

**Note:**  Before this PR get merged, `secrets.GITHUB_TOKEN` and `secrets.TAP_GITHUB_TOKEN` need to be set in the env


phileba (approved):
@ivanvanderbyl can we update the secret token then get it to go?

PR #268, opened 2020-09-09
Features/enhance verbose mode report

### LINKED ISSUE #266: [Enhancement] Add more details in Verbose mode terminal
https://tricentisflood.atlassian.net/browse/FLO-1827

The logs in terminal while running with Verbose mode should include:
- The beginning (which will be turned into ending after finished) of each step
- The beginning (which will be turned into ending after finished) of each hook (before..., after...)
- Name of the action within each step, while that action is being executed
- Name of the conditional steps and skipped steps even if those steps are not executed
- Console (log/error/warning/debug/info) messages

While the test is running
image

And after it finished
image

ai-flood:
resolve #266

### DIFF
--- a/packages/core/src/runtime/Test.ts
+++ b/packages/core/src/runtime/Test.ts
@@ -220,7 +220,7 @@ export default class Test implements ITest {
 					if (result) {
 						this.failed = false
 					} else {
-						throw Error('recovery step -> failed')
+						throw Error('step -> failed')
 					}
 				}
 			})
--- a/packages/core/src/runtime/Test.ts
+++ b/packages/core/src/runtime/Test.ts
@@ -181,13 +181,7 @@ export default class Test implements ITest {
 			}
 
 			debug('running hook function: beforeAll')
-			await this.runHookFn(
-				this.hook.beforeAll,
-				browser,
-				testDataRecord,
-				HookType.beforeAll,
-				testObserver,
-			)
+			await this.runHookFn(this.hook.beforeAll, browser, testDataRecord, testObserver)
 
 			debug('running steps')
 			await stepIterator.run(async (step: Step) => {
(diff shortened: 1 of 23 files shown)


packages/report/src/reporters/Verbose.ts
```diff
@@ -33,31 +33,54 @@ export class VerboseReporter implements IReporter {
 	): void {
 		const stepName = 'Step ' + (subtitle ? `'${label}' (${subtitle})` : `'${label}'`)
 		const beforeRunStepMessage = `${stepName} is running ...`
+		const beforeRunHookMessage = chalk.grey(`${label} is running ...`)
+		const afterRunHookMessage = `${chalk.green.bold('✔')} ${chalk.grey(`${label} finished`)}`
 		let message = ''
 		switch (stage) {
```
HieuLeKMS, line 39: I think we should remove the unhandled cases in this `switch`

packages/core/src/runtime/Test.ts
```diff
@@ -184,16 +181,19 @@ export default class Test implements ITest {
...
 
 			debug('running hook function: beforeAll')
-			await this.runHookFn(this.hook.beforeAll, browser, testDataRecord)
+			await this.runHookFn(
+				this.hook.beforeAll,
+				browser,
+				testDataRecord,
+				HookType.beforeAll,
+				testObserver,
+			)
```
phileba, line 190: Just a sugesstion, I think we do not need arg `HookType.beforeAll`, can we get the `type` from the `this.hook.beforeAll` by adding property in Hookbase? 
```suggestion
			await this.runHookFn(
				this.hook.beforeAll,
				browser,
				testDataRecord,
				testObserver,
			)
```

phileba (approved):

PR #274, opened 2020-09-16
fix: fix consoleFilter doesnot work

ai-flood:


phileba (approved):
LGTM, plz fix test failed then we can 🚀

PR #284, opened 2020-09-22
fix: some bugs on Element API

HieuLeKMS:
Fix:
- `ElementHandle.uploadFile()` has been implemented but is not exposed in the API
- `Browser.wait()` returns value of type boolean only
Enhance:
- Show Element version in log when running tests


honglax (approved):
LGTM 🐑

phileba (approved):

PR #285, opened 2020-09-23
fix: handle fail exit code when running on local only

HieuLeKMS:
Fix: On Grids, tests stop running if there's a failed step, instead of skip to the next iteration

### DIFF
--- a/packages/cli/src/cmd/run.ts
+++ b/packages/cli/src/cmd/run.ts
@@ -184,14 +184,7 @@ async function runTestScript(args: RunCommonArguments): Promise<void> {
 		opts.testCommander = makeTestCommander(file)
 	}
 
-	try {
-		await runCommandLine(opts)
-	} catch (err) {
-		console.log('Element exited with error')
-		console.error(err)
-		const exitCode = args['fail-status-code']
-		process.exit(exitCode)
-	}
+	await runCommandLine(opts)
 }
 
 async function runTestScriptWithConfiguration(args: RunCommonArguments): Promise<void> {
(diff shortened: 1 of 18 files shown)


honglax (approved):
LGTM

packages/cli/src/cmd/run.ts
```diff
@@ -185,7 +184,14 @@ async function runTestScript(args: RunCommonArguments): Promise<void> {
 		opts.testCommander = makeTestCommander(file)
 	}
 
-	await runCommandLine(opts)
+	try {
+		await runCommandLine(opts)
+	} catch (err) {
+		console.log('Element exited with error')
+		console.error(err)
+		const exitCode = args['fail-status-code']
```
phileba, line 192: do we have a default `exitCode`?

HieuLeKMS: `fail-status-code`'s default value is 1
image

phileba: I think the default should be `0` cause the issue when running on flood, I tried with a simple script with a timeout error and the exit code is `1`

HieuLeKMS: The default exit code when a test failed should be 1, to indicate that there was an error. Beside, I fixed it so that a test won't stop even if it failed when running on a grid, `fail-status-code` won't be used in that case

phileba (approved):

HieuLeKMS:
Duplicate of https://github.com/flood-io/element/pull/292. Closing this PR

PR #297, opened 2020-10-01
[Element 1.4] Implement new API browser.scrollBy() and browser.scrollTo()

honglax:
Implementing new API for Element.

**browser.scrollBy(x: number, y: number, [,option])**
```
- x: How many pixels to scroll by, along the x-axis (horizontal). Positive values will scroll to the right, while negative values will scroll to the left. 
- x can take window.innerWidth as a special value.
- y: How many pixels to scroll by, along the y-axis (vertical). Positive values will scroll down, while negative values scroll up.
- y can take window.innerHeight as a special value.
- option: is an object containing ScrollToOptions.

E.g.: 
await browser.scrollBy(100, 0, { behavior: 'smooth' })
await browser.scrollBy('window.innerWidth', 0, { behavior: 'smooth' })
await browser.scrollBy(0, 'window.innerHeight', { behavior: 'smooth' })

const innerHeight: number = await browser.evaluate(() => window.innerHeight)
const innerWidth: number = await browser.evaluate(() => window.innerWidth)
await browser.scrollBy(innerWidth, innerHeight, { behavior: 'smooth' })
```

**browser.scrollTo(target: Point, [,option])**
```
- point is an array of x (number) and y (number) co-ordinate
- options is an object containing ScrollToOptions.

E.g.: await browser.scrollTo([69, 96], { behavior: 'smooth', block: 'nearest' })
```

**browser.scrollTo(target: ElementHandle, [,option])**
```
E.g.: 
const title = await browser.findElement(By.css('#title'))
await browser.scrollTo(title, { behavior: 'smooth', block: 'center', inline: 'center' })
```

**browser.scrollTo(target: Locator, [,option])**
`E.g.: await browser.scrollTo(By.visibleText('submit'), { behavior: 'smooth' })`

**browser.scrollTo(ScrollDirection: string, [,option])**
```
ScrollDirection can be one of the following: top, bottom, left, or right 
E.g.: await browser.scrollTo("top", { behavior: 'smooth' })
```

### DIFF
--- a/examples/flood-element-challenge.ts
+++ b/examples/flood-element-challenge.ts
@@ -1,0 +1,427 @@
+import {
+	step,
+	TestSettings,
+	By,
+	Until,
+	Key,
+	Browser,
+	beforeAll,
+	beforeEach,
+	afterEach,
+	afterAll,
+} from '@flood/element'
+import assert from 'assert'
+
+export const settings: TestSettings = {
+	loopCount: 1,
+	waitUntil: 'visible',
+	clearCache: false,
+	clearCookies: false,
+	disableCache: false,
+	waitTimeout: '60s',
+	screenshotOnFailure: true,
+	viewport: { width: 1920, height: 1080 },
+	launchArgs: ['--window-size=1920,1080'],
+	stepDelay: '1s',
+	actionDelay: '500ms',
+}
+
+const URL = 'https://flood-element-challenge.vercel.app/'
+
+const getNumOfPages = async (browser: Browser): Promise<number> => {
+	return await browser.evaluate(() => {
+		const nav = document.querySelector('nav[aria-label="pagination navigation"]')
+		const items = nav.getElementsByTagName('li')
+		return items.length === 2 ? 1 : items.length - 2
+	})
+}
+
+const scrollToTop = async (browser: Browser): Promise<void> => {
+	await browser.scrollTo('top', { behavior: 'smooth' })
+}
+
+const adjustPriceRange = async (
+	browser: Browser,
+	targetMinPrice: number,
+	targetMaxPrice: number,
+): Promise<void> => {
+	const priceRange = await browser.findElement(By.css('p[data-test-range]'))
+	const priceSlider = await browser.findElement(By.css('span[data-test-slider]'))
+
+	await priceRange.click()
+
+	const sliderLocation = await priceSlider.location()
+	const sliderWidth = (await priceSlider.size()).width
+
+	const maxLocation = sliderLocation.x + 1950 / (1950 / sliderWidth)
+
+	const minPriceLocation = sliderLocation.x + (targetMinPrice - 50) / (1950 / sliderWidth)
+	const maxPriceLocation = sliderLocation.x + (targetMaxPrice - 50) / (1950 / sliderWidth)
+
+	await browser.mouse.drag([maxLocation, sliderLocation.y], [maxPriceLocation, sliderLocation.y])
+
+	const priceRangeText = await priceRange.text()
+	let currentMaxPrice = parseInt(
+		priceRangeText
+			.split('-')[1]
+			.trim()
+			.replace('$', ''),
+	)
+
+	while (currentMaxPrice !== targetMaxPrice) {
+		if (currentMaxPrice > targetMaxPrice) {
+			await browser.press(Key.ARROW_LEFT)
+			currentMaxPrice--
+		} else {
+			await browser.press(Key.ARROW_RIGHT)
+			currentMaxPrice++
+		}
+	}
+
+	await browser.mouse.drag(
+		[sliderLocation.x, sliderLocation.y],
+		[minPriceLocation, sliderLocation.y],
+	)
+
+	const newPriceRangeText = await priceRange.text()
+	let currentMinPrice = parseInt(
+		newPriceRangeText
+			.split('-')[0]
+			.trim()
+			.replace('$', ''),
+	)
+
+	while (currentMinPrice !== targetMinPrice) {
+		if (currentMinPrice > targetMinPrice) {
+			await browser.press(Key.ARROW_LEFT)
+			currentMinPrice--
+		} else {
+			await browser.press(Key.ARROW_RIGHT)
+			currentMinPrice++
+		}
+	}
+}
+
+const minimizePopup = async (browser: Browser): Promise<void> => {
+	const minimizeBtn = await browser.findElement(By.xpath('//button[@data-test-minimize]'))
+	await browser.click(minimizeBtn)
+	await browser.wait(0.5)
+}
+
+const maximizePopup = async (browser: Browser): Promise<void> => {
+	const maximizeBtn = await browser.findElement(By.xpath('//button[@data-test-maximize]'))
+	await browser.click(maximizeBtn)
+	await browser.wait(0.5)
+}
+
+const getChallangeText = async (browser: Browser): Promise<string> => {
+	const modal = await browser.findElement(By.id('challenges-popup'))
+	const challengeText = await (await modal.findElement(By.xpath('//h6[@data-test-title]'))).text()
+
+	return challengeText.split(' ')[1].split('/')[0]
+}
+
+export default () => {
+	beforeAll(async browser => {
+		await browser.visit(URL)
+
+		const pageTextVerified = By.css('h3')
+		await browser.wait(Until.elementIsVisible(pageTextVerified))
+
+		await browser.press(Key.TAB)
+		await browser.press(Key.ENTER)
+
+		// const takeTheChallengeBtn = await browser.findElement(By.visibleText('TAKE THE CHALLENGE'))
+		// await browser.click(takeTheChallengeBtn)
+
+		await browser.takeScreenshot()
+		const modal = By.css('#challenges-popup')
+		await browser.wait(Until.elementIsVisible(modal))
+	})
+
+	beforeEach(async browser => {
+		const challenge = await getChallangeText(browser)
+		if (challenge && parseInt(challenge) <= 8) {
+			const nextChallengeBtn = By.visibleText('CHECK')
+			await browser.wait(Until.elementLocated(nextChallengeBtn))
+		}
+	})
+
+	afterEach(async browser => {
+		await browser.takeScreenshot()
+		const modal = await browser.findElement(By.id('challenges-popup'))
+
+		const checkBtn = await modal.findElement(By.visibleText('CHECK'))
+		checkBtn?.click()
+
+		const nextBtn = By.visibleText('NEXT')
+		await browser.wait(Until.elementLocated(nextBtn))
+		const nextBtnEl = await modal.findElement(By.visibleText('NEXT'))
+		nextBtnEl.click()
+	})
+
+	afterAll(async browser => {
+		const finalMessage = By.xpath('//h5[@data-test-message]')
+		await browser.wait(Until.elementLocated(finalMessage))
+		const finalMsgText = await (await browser.findElement(finalMessage)).text()
+		assert.strictEqual(
+			finalMsgText,
+			'🎉hooray!',
+			'The final message should be correct after you finished 8 steps',
+		)
+		await browser.wait(5)
+	})
+
+	step('Test: Challenge 1', async browser => {
+		const challenge = await getChallangeText(browser)
+		assert.strictEqual(challenge, '1', 'The challenge should be challenge number 1')
+
+		const h2El = await browser.findElement(By.css('h2'))
+		const discountText = await h2El.text()
+		const discountNumber = discountText.split(' ').filter(word => word.includes('%'))[0]
+
+		const modalEl = await browser.findElement(By.css('#challenges-popup'))
+		const labels = await modalEl.findElements(By.tagName('label'))
+		await Promise.all(
+			labels.map(async label => {
+				const correctElText = await label.text()
+				if (correctElText === discountNumber) {
+					await label.click()
+				}
+			}),
+		)
+	})
+
+	step('Test: Challenge 2', async browser => {
+		const challenge = await getChallangeText(browser)
+		assert.strictEqual(challenge, '2', 'The challenge should be challenge number 2')
+
+		const modalEl = await browser.findElement(By.id('challenges-popup'))
+		const keyword = await (await modalEl.findElement(By.id('challenge-2-category')))?.text()
+		if (keyword?.toLowerCase() !== 't-shirt') {
+			const targetButton = await browser.findElement(By.xpath(`//button[@value="${keyword}"]`))
+			await targetButton.click()
+		}
+
+		const productCard = By.css('div[data-test-card]')
+		await browser.wait(Until.elementsLocated(productCard))
+
+		const label = await modalEl.findElement(
+			By.js(() => {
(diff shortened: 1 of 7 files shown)


phileba:
`browser.scrollTo([69, 96], { behavior: "smooth" })` @sonnguyenmau @honglax do you think we need to support point array, it will scroll point by point in this array, maybe we should have an option like `scrollDelay` 🤔

sonnguyenmau:
Hi a @phileba , I don't think we need to support this. The user can always use 2 separate scrolls to achieve the same result. I think we should support 2 special params, which are window.innerHeight and window.innerWidth. What do you think?

phileba:
> Hi a @phileba , I don't think we need to support this. The user can always use 2 separate scrolls to achieve the same result. I think we should support 2 special params, which are window.innerHeight and window.innerWidth. What do you think?

okay Sơn, so how to support this?

sonnguyenmau:
instead of using only number, users can input window.innerHeight or window.innerWidth as special values. Don't worry a Phi, @honglax will handle this

phileba:
> instead of using only number, users can input window.innerHeight or window.innerWidth as special values. Don't worry a Phi, @honglax will handle this

okay Sơn, I believe him :v

packages/core/src/page/types.ts
```diff
@@ -202,3 +202,6 @@ export interface TargetLocator {
 	 */
 	frame(id: number | string | ElementHandle): Promise<void>
 }
+
+// Using these type below for API scrollTo()
```
HieuLeKMS, line 206:
```suggestion
// Using the type below for API scrollTo()
```

packages/core/src/runtime/Browser.ts
```diff
@@ -600,4 +601,131 @@ export class Browser<T> implements BrowserInterface {
...
+
+	@addCallbacks()
+	public async scrollTo(
+		target: Locator | ElementHandle | Point | ScrollDirection,
+		scrollOptions?: ScrollIntoViewOptions,
+	): Promise<void> {
+		const behavior = scrollOptions?.behavior ?? 'auto'
+
+		if (!this.isCorrectScrollBehavior(behavior)) {
+			throw new Error('The input behavior is not correct (Must be "auto" or "smooth").')
```
HieuLeKMS, line 638: It's not really matter but `throw new Error(...)` & `throw Error(...)` are equivalent

packages/core/src/page/types.ts
honglax: Resolved.

packages/core/src/runtime/Browser.ts
honglax: So should we use `throw Error(...)` instead of `throw new Error(...)`?

HieuLeKMS: I think either is fine. I just informed you in case you didn't know.

honglax: Got it. Thanks for your advice.

packages/core/src/runtime/Browser.ts
```diff
@@ -600,4 +601,143 @@ export class Browser<T> implements BrowserInterface {
 	public async close(): Promise<void> {
 		await this.client.browser.close()
 	}
+
+	// Check the target has type Locator or not
+	private isLocator(target: Locator | ElementHandle | Point | ScrollDirection): target is Locator {
+		return !Array.isArray(target) && (target as Locator).find !== undefined
```
phileba, line 607: This check is not safe I think, it will be an issue if others target (e.g: ElementHandle) has method `find`

packages/core/src/runtime/Browser.ts
```diff
@@ -600,4 +601,143 @@ export class Browser<T> implements BrowserInterface {
...
+	}
+
+	// Check the target has type ElementHandle or not
+	private isElementHandle(
+		target: Locator | ElementHandle | Point | ScrollDirection,
+	): target is ElementHandle {
+		return (target as ElementHandle).element !== undefined
```
phileba, line 614: the same as above, I think we should have a method like `type` for Locator and ElementHandle, then it will return exactly which type of the target.

honglax: So should I add a method `type` for **Locator** and **ElementHandle**? Is it a breaking change or not?

packages/core/src/runtime/Browser.ts
```diff
@@ -600,4 +601,143 @@ export class Browser<T> implements BrowserInterface {
...
+	}
+
+	private isPoint(target: Locator | ElementHandle | Point | ScrollDirection): target is Point {
+		return (
+			Array.isArray(target) &&
+			target.length === 2 &&
+			typeof target[0] === 'number' &&
+			typeof target[1] === 'number'
```
phileba, line 622: is it enough?
```suggestion
			Array.isArray(target) &&
			typeof target[0] === 'number' &&
			typeof target[1] === 'number'
```

packages/core/src/runtime/Browser.ts
```diff
@@ -600,4 +601,143 @@ export class Browser<T> implements BrowserInterface {
...
+		)
+	}
+
+	private isCorrectScrollBehavior(behavior: string): behavior is ScrollBehavior {
+		return behavior === 'auto' || behavior === 'smooth'
```
phileba, line 627:
```suggestion
		return ['auto', 'smooth'].includes(behavior)
```

packages/core/src/runtime/Browser.ts
honglax: To prevent the end-user using `[1, 2, 'hello world']` or something like this.

packages/core/src/runtime/Browser.ts
honglax: Resolved.

packages/core/src/runtime/Browser.ts
```diff
@@ -600,4 +601,143 @@ export class Browser<T> implements BrowserInterface {
...
+		let top = 0
+		let left = 0
+		const [_scrollHeight, _currentTop, _scrollWidth] = await this.page.evaluate(() => [
+			document.body.scrollHeight,
+			window.pageYOffset || document.documentElement.scrollTop,
+			document.body.scrollWidth,
+		])
+
+		if (this.isPoint(target)) {
+			// target is Point
```
phileba, line 653: the function already told this target is a Point so we don't need this comment I think

packages/core/src/runtime/Browser.ts
```diff
@@ -600,4 +601,143 @@ export class Browser<T> implements BrowserInterface {
...
+			top = target[1]
+			left = target[0]
```
phileba, line 655:
```suggestion
			[left, top] = target
```

packages/core/src/runtime/Browser.ts
honglax: Resolved

packages/core/src/runtime/Browser.ts
phileba: in this case, the target is Point, so it should be `[number, number]`, the compile will fail if the input is `[1, 2, 'hello world']` 🤔

honglax: I want to check the length because I want to throw a friendly error message to the end-user if they input the incorrect type of param. The compile will fail but maybe the end-user will not get a friendly error message.

phileba: okay

packages/core/src/runtime/Browser.ts
phileba: please be careful when you throw an Error like this, it will stop the test when running on Flood.

packages/core/src/runtime/Browser.ts
```diff
@@ -600,4 +601,143 @@ export class Browser<T> implements BrowserInterface {
...
+					top = _currentTop
+					left = _scrollWidth
+					break
+				default:
+					throw new Error(
+						'The input target is not Locator or ElementHandle or Point or Scroll Direction.',
+					)
+			}
+		} else if (this.isLocator(target) || this.isElementHandle(target)) {
+			// target is Locator or ElementHandle
```
phileba, line 681: 🔥

packages/core/src/runtime/Browser.ts
```diff
@@ -600,4 +601,143 @@ export class Browser<T> implements BrowserInterface {
...
+			const targetEl = await this.findElement(target)
+			await targetEl.element.evaluate(
+				(el, scrollOptions: ScrollIntoViewOptions) => {
+					el.scrollIntoView(scrollOptions)
+				},
+				{ behavior, block, inline },
+			)
+			return
```
phileba, line 689: for this block, we return when complete the action so I think we should move this to the top as an early return pattern.

packages/core/src/runtime/Browser.ts
honglax: Resolved.

examples/flood-element-challenge.ts
```diff
@@ -0,0 +1,427 @@
...
+		const numOfPage = await getNumOfPages(browser)
+
+		for (let x = 1; x <= numOfPage; x++) {
+			const productCards = await browser.findElements(By.xpath('//div[@data-test-card]'))
+			for (let y = 0; y < productCards.length; y++) {
+				// await browser.scrollTo(productCards[y])
+				// await browser.evaluate(y => {
+				// 	const el = document.querySelectorAll('div[data-test-card]')[y]
+				// 	el.scrollIntoView({ block: 'nearest' })
+				// }, y)
```
phileba, line 310: 🔥

examples/flood-element-challenge.ts
```diff
@@ -0,0 +1,427 @@
...
+				await browser.click(anotherAddToCardBtn)
+
+				// Using Mouse.move(), mouse.down() and mouse.up() to close the modal
+				await browser.mouse.move(1, 1)
+				await browser.mouse.down()
+				await browser.mouse.up()
+
+				// Or you can click into the close button in the modal
+				// const closeModalBtnEl = await browser.findElement(By.css('button[data-test-product-detail-modal-close]'))
+				// await closeModalBtnEl.click()
```
phileba, line 333: 🔥

packages/core/src/runtime/Browser.ts
phileba: I don't want to change the interface so find another solution with a smaller scope

examples/flood-element-challenge.ts
honglax: Resolved.

examples/flood-element-challenge.ts
honglax: Resolved.

packages/core/src/runtime/Browser.ts
honglax: Resolved.

packages/core/src/runtime/Browser.ts
honglax: Got it.

HieuLeKMS: You might want to try the operator `instanceof`

phileba: yep, so we can distinguish `Locator` and `ElementHandle` by:
```ts
import { BaseLocator } from '../page/Locator'
import { ElementHandle as TargetElementHandle } from '../page/Locators'
...
if (target instanceof BaseLocator){
// locator
} else if (target instanceof TargetElementHandle) {
// element handle
}

Comment (honglax) [reply packages/core/src/runtime/Browser.ts:614]: Unfortunately, `Locator` and `ElementHandle` just are `Interface` hence we can not use `instanceof` with them.

Comment (phileba) [reply packages/core/src/runtime/Browser.ts:614]: Yep I know but I'm talking about class `BaseLocator` and `ElementHandle` :slap:

Comment (HieuLeKMS) [reply packages/core/src/runtime/Browser.ts:614]: You can import and check with the classes that implement those interfaces as @phileba suggested

Comment (honglax) [reply packages/core/src/runtime/Browser.ts:614]: Got it.

File: packages/core/src/runtime/Browser.ts (hunk 10 above)
Comment (honglax) [reply packages/core/src/runtime/Browser.ts]: Resolved.

File: packages/core/src/runtime/Browser.ts (hunk 4 above)
Comment (honglax) [reply packages/core/src/runtime/Browser.ts]: Resolved.

File: packages/core/src/runtime/Browser.ts (hunk 3 above)
Comment (honglax) [reply packages/core/src/runtime/Browser.ts]: Resolved.

File: packages/core/src/runtime/Browser.ts [hunk 13]
```diff
@@ -600,4 +603,141 @@ export class Browser<T> implements BrowserInterface {
...
+		let top = 0
+		let left = 0
+		const [_scrollHeight, _currentTop, _scrollWidth] = await this.page.evaluate(() => [
+			document.body.scrollHeight,
+			window.pageYOffset || document.documentElement.scrollTop,
+			document.body.scrollWidth,
+		])
+
+		if (this.isLocator(target) || this.isElementHandle(target)) {
+			const targetEl = await this.findElement(target)
```
phileba, line 655: if target already is an `ElementHandle`, do we need to `findElement` again?

honglax: `ElementHandle.findElement()` returns an ElementHandle hence I use it to avoid using too much if-else.

phileba: It will call `Browser.findElement`, then it will try to cast the `ElementHandle` to `Locator` in `locatableToLocator` function, then call `locator.find` to return an `ElementHandle`. Luckily the ElementHandle itself already has the method `find` too so it works but the logic is incorrect and we might have an issue with this in the future 🤔

honglax: So I'll use `findElement()` if the **target Element** is `Locator` to prevent any issue.

phileba: I think so, it's easy to maintain later

honglax: Thanks for your advice.

packages/core/src/runtime/Browser.spec.ts
```diff
@@ -228,4 +237,124 @@ describe('Browser', () => {
...
+			docClientHeight = await browser.evaluate(() => document.documentElement.clientHeight)
+			docClientWidth = await browser.evaluate(() => document.documentElement.clientWidth)
+		})
+
+		afterEach(async () => {
+			await browser.scrollTo('top')
+		})
+		test('scrollTo', async () => {
+			await browser.scrollTo('bottom', { behavior: 'smooth' })
+			await browser.wait(2)
```
HieuLeKMS, line 284: What are these `browser.wait(second)` for? I think these will increase a good amount of time for running unit tests. Should we remove them?

honglax: When we use `scroll` with `behavior`, it will take time to scroll to the target position. Hence, I need to wait to make sure the scroll action has finished.

HieuLeKMS (approved):

honglax:
Hi @HieuLeKMS, I've refactored the unit-test file to remove the redundant line of code. I've also added a HTML file to using for test the API scrollTo() and scrollBy().

HieuLeKMS:
LGTM. 🚢 when you're 😃

PR #299, opened 2020-10-05
Feature: export test result as HTML

HieuLeKMS:


phileba (approved):
Looks great, please resolve the conflict then we can get it go 🚢

PR #303, opened 2020-10-06
[Element 1.4] Edit publish config to specify the exact version of updated dependencies

honglax:
Add a flag `--exact` when publishing a new version of stable version and beta version to specify updated dependencies in updated packages **exactly** (with no punctuation), instead of as semver compatible **(with a ^)**.

More specifically, please take a look at [documentation of lerna](https://github.com/lerna/lerna/tree/master/commands/version#--exact).

### DIFF
--- a/.github/workflows/lerna-beta.yml
+++ b/.github/workflows/lerna-beta.yml
@@ -37,7 +37,7 @@ jobs:
       - name: 'Build all'
         run: yarn build
       - name: 'Publish beta to NPM'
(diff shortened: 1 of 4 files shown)


phileba:
@honglax do we need to update the package.json file (remove the `^`) in other packages?

honglax:
> @honglax do we need to update the package.json file (remove the `^`) in other packages?

Nope, we don't need to do this. Lerna will do it for us every time we publish new version.

phileba (approved):

PR #313, opened 2020-10-21
[Element v1.4] Prevent Lerna auto-bumping version when modifying docs (*.md & packages/docs/**)

honglax:
Edit file `lerna.json`:

- Add option `ignoreChanges: ["*.md", "packages/docs/**"]` into `command.publish`.
- Add `**/*md` & `packages/docs/**` to `ignoreChanges`.

For more information, please take a look at [Lerna.json](https://github.com/lerna/lerna#lernajson) and [lerna version --ignore-change](https://github.com/lerna/lerna/tree/master/commands/version#--ignore-changes)

### DIFF
--- a/.github/workflows/lerna-beta.yml
+++ b/.github/workflows/lerna-beta.yml
@@ -37,7 +37,7 @@ jobs:
       - name: 'Build all'
         run: yarn build
       - name: 'Publish beta to NPM'
-        run: yarn lerna publish --conventional-prerelease --dist-tag beta --yes --preid beta --include-merged-tags --force-publish --exact
+        run: yarn lerna publish --conventional-prerelease --dist-tag beta --yes --preid beta --include-merged-tags --force-publish --exact --ignore-changes '**/*.md'
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
       - name: 'Publish beta to Homebrew'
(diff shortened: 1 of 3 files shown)


.github/workflows/lerna-beta.yml
```diff
@@ -37,7 +37,7 @@ jobs:
       - name: 'Build all'
         run: yarn build
       - name: 'Publish beta to NPM'
-        run: yarn lerna publish --conventional-prerelease --dist-tag beta --yes --preid beta --include-merged-tags --force-publish --exact
+        run: yarn lerna publish --conventional-prerelease --dist-tag beta --yes --preid beta --include-merged-tags --force-publish --exact --ignore-changes '**/*.md'
```
HieuLeKMS, line 40: Do we need to provide `--ignore-changes` flag here (and in `lerna-stable.yml` as well)? Since you have already had the `ignoreChanges` property in `lerna.json`

honglax: I need to make sure that the flag `--ignore-changes` works well when publishing stable and beta version.

lerna.json
```diff
@@ -8,7 +8,8 @@
 			"exact": true
 		},
 		"publish": {
-			"conventionalCommits": true
+			"conventionalCommits": true,
+			"ignoreChanges": ["*.md"]
```
HieuLeKMS, line 12: Same as my above comment. Do we really need this?

phileba: yep, I think we don't need it because we already have it in the `ignoreChanges` in the root of the `lerna.json` configuration.

honglax: Roger that.

phileba (approved):
after resolving the comment then we can 🚢

PR #321, opened 2020-10-30
[Element 1.4] Introduce `browser.getUrl()` as new API

honglax:
**browser.getUrl()**

**Description**: Return the URL of the current page as a string.

**Params**: none

**Return**: current URL as a string

### DIFF
--- a/packages/core/src/runtime/Browser.ts
+++ b/packages/core/src/runtime/Browser.ts
@@ -739,8 +739,7 @@ export class Browser<T> implements BrowserInterface {
 		)
 	}
 
-	@addCallbacks()
-	public async getUrl(): Promise<string> {
+	public getUrl(): string {
 		return this.page.url()
 	}
 }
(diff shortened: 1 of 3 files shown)


packages/core/src/runtime/Browser.ts
```diff
@@ -738,4 +738,9 @@ export class Browser<T> implements BrowserInterface {
 			behavior,
 		)
 	}
+
+	@addCallbacks()
```
HieuLeKMS, line 742: I don't think we need `addCallbacks` for `getUrl`, since it's not an action on browser. And without it, `getUrl` would return a string instead of a Promise, that I think would be better.

honglax: I got your idea.

honglax: Hi @HieuLeKMS, please review the changes and let me know if you have any concern. Thank you.

HieuLeKMS (approved):

PR #327, opened 2020-11-03
[Element 2.0] Create consistency between browser's actions of script and reporter

honglax:
I've created new functions without the decorator, and call it from the original function. I also edited the decorator `autoWaitUntil`.

### DIFF
--- a/packages/core/src/runtime/Browser.ts
+++ b/packages/core/src/runtime/Browser.ts
@@ -574,7 +574,6 @@ export class Browser<T> implements BrowserInterface {
 
 	private async findElementWithoutDecorator(locatable: NullableLocatable): Promise<ElementHandle> {
 		const locator = locatableToLocator(locatable, 'browser.findElement(locatable)')
-
 		debug('locator %o', locator)
 
 		const maybeElement = await locator.find(this.page, undefined, this.target)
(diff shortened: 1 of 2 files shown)


packages/core/src/runtime/Browser.ts
```diff
@@ -607,4 +567,58 @@ export class Browser<T> implements BrowserInterface {
...
+		const locator = locatableToLocator(locatable, 'browser.findElement(locatable)')
+
+		debug('locator %o', locator)
+
+		const maybeElement = await locator.find(this.page, undefined, this.target)
+		if (!maybeElement) {
+			throw toLocatorError(locatable, 'browser.findElement()')
+		}
+		const element = maybeElement as ElementHandle
+
```
phileba, line 585: should remove this empty line

honglax: Got it.

honglax: Hi a @phileba, please check the change and let me know if you have any concern. Thank you.

phileba: this is small, you can change and mark resolved by yourself.

phileba (approved):

PR #328, opened 2020-11-05
[element 1.4] support api getMimeType of file

phileba:


packages/cli/src/utils/ConsoleReporter.ts
```diff
@@ -89,7 +89,7 @@ cause.stack: ${detail.causeStack}`)
 
 	testScriptConsole(method: string, message?: any, ...optionalParams: any[]): void {
 		debug('testScriptConsole', method, message)
-		if (method === 'log') method = 'info'
+		if (['log', 'clear'].includes(method)) method = 'info'
```
HieuLeKMS, line 92: I don't get what's your intention here, can you explain it to me?

honglax: me too

packages/core/src/runtime/Browser.ts
```diff
@@ -661,7 +666,7 @@ export class Browser<T> implements BrowserInterface {
 		}
 
 		if (this.isPoint(target)) {
-			[left, top] = target
+			;[left, top] = target
```
honglax, line 669: We should remove `;`

packages/cli/src/utils/ConsoleReporter.ts
phileba: we attached a console event to the page and handle this in the reporter
```
await this.networkRecorder.attachEvent('console', msg => {
	const msgType = msg.type() === 'warning' ? 'warn' : msg.type()
	if (this.consoleFilters.length === 0 || !this.consoleFilters.includes(msgType)) {
		this.reporter.testScriptConsole(msg.type(), msg.text())
	}
})
```
this page https://4hzsj.csb.app/ has `msg.type() = clear` and made the element crashed

packages/core/src/runtime/Browser.ts
phileba: it caused prettier lol

packages/cli/src/utils/ConsoleReporter.ts
HieuLeKMS: But I think we should return early if `method === 'clear'`, instead of change it from 'clear' to 'info'. Since `console.clear()` is not a logging method

packages/core/src/runtime/Browser.ts
honglax: Would it ok if we put it there?

HieuLeKMS: It was added by prettier auto-formatting, and there won't be a problem with it so we should leave it there

packages/cli/src/utils/ConsoleReporter.ts
phileba: early return is ok, I just think if some case we need to know the message of this method.

HieuLeKMS: `console.clear()` doesn't have a message as parameter though

HieuLeKMS (approved):

PR #329, opened 2020-11-06
Update core Puppeteer to the latest version

honglax:
Work in progress...


packages/core/package.json
```diff
@@ -48,7 +48,7 @@
 		"ksuid": "^1.2.0",
 		"micromatch": "^4.0.2",
 		"ms": "^2.1.2",
-		"puppeteer": "^2.1.1",
+		"puppeteer": "^5.4.1",
```
HieuLeKMS, line 51: We only need puppeteer `v3.1`, right? Is it really okay to go all the way to `v5.4.1`?

honglax: Per our discussion yesterday, a @phileba decided to check with the latest version of Puppeteer. After checking, fixing some minor issues, and running the test suite with Jest, there are no issues happen. Therefore, I choose `v.5.4.1` instead of `v.3.1`

phileba: we need to run the flood with the element latest puppeteer to make sure everything works well.

honglax: I think so. But we need to have the canary channel on NPM to make sure it will not break anything.

honglax:
Close this PR due to error

PR #338, opened 2020-11-17
[element 2.0] Support browserLaunchOptions in TestSetting

phileba:
resolve https://tricentisflood.atlassian.net/browse/FLO-2004


packages/core/src/page/types.ts
```diff
@@ -1,11 +1,7 @@
 import { ElementHandle as PElementHandle, Frame, Page } from 'playwright'
 import { Point } from './Point'
 
-export enum BROWSER_TYPE {
-	CHROME = 'chromium',
-	FIREFOX = 'firefox',
-	WEBKIT = 'webkit',
-}
+export type BrowserType = 'chromium' | 'firefox' | 'webkit'
```
ivanvanderbyl, line 4: Should this include 'edge' ?

phileba: they are 3 types were supported by playwright, to run with Edge, the user should be defined in `TestSetting`
```ts
	loopCount: 1,
	browserLaunchOption: {
		executablePath: '/Applications/Microsoft Edge.app/Contents/MacOS/Microsoft Edge',
		headless: false,
	},
}
``` 
we can not handle the path of the Edge in cross-platform cause of some permission, so let the user do it.

packages/core/src/runtime/Test.spec.ts
```diff
@@ -5,10 +5,9 @@ import Test from './Test'
 import { EvaluatedScript } from './EvaluatedScript'
 import { join } from 'path'
 import { EventEmitterReporter } from '@flood/element-report'
-import { ConcreteTestSettings, normalizeSettings, TestSettings } from './Settings'
```
honglax, line 8: Is TestSettings redundant?

phileba: yep @honglax

honglax (approved):
LGTM 🚀

PR #353, opened 2020-11-24
[Element Beta] Update core Puppeteer to the latest version

honglax:


.github/workflows/lerna-beta.yml
```diff
@@ -40,5 +40,3 @@ jobs:
         run: yarn lerna publish --conventional-prerelease --dist-tag beta --yes --preid beta --include-merged-tags --force-publish --exact --ignore-changes '**/*.md' 'packages/docs/**'
         env:
           NODE_AUTH_TOKEN: ${{ secrets.NPM_TOKEN }}
-      - name: 'Publish beta to Homebrew'
-        run: yarn publish:brew
```
HieuLeKMS, line 44: you might want to remove this in other workflows as well, it's not working anyway.

honglax: I've removed it on beta and canary.

HieuLeKMS: You should remove it in the stable workflow as well, it's not working and will cause errors when you run it, also, it might cause confusion later on because there're differences in steps between workflows. We'll fix it for all workflows at once when we'll have found a solution for it.

honglax: Got it

HieuLeKMS (approved):

PR #359, opened 2020-11-24
[Element 2.0] Time unit consistency and improvement

phileba:


honglax (approved):

packages/core/src/runtime/Settings.ts
```diff
@@ -301,46 +308,57 @@ export type ConcreteTestSettings = Required<TestSettings>
...
-		convertedDuration = settings.duration
+	if (settings.duration) {
+		if (typeof settings.duration === 'string') {
+			convertedDuration = ms(settings.duration)
+		} else {
+			convertedDuration = settings.duration
+		}
+		settings.duration = convertedDuration
 	}
-	settings.duration = convertedDuration > 0 ? convertedDuration : -1
```
HieuLeKMS, line 343: Isn't `settings.duration` is -1 by default? It might possibly be `undefined` like this right?

packages/cli/src/cmd/run.ts
```diff
@@ -83,6 +83,7 @@ const cmd: CommandModule = {
 				},
 			)
 		}
+		process.exit(0)
```
HieuLeKMS, line 86: what are these `process.exit(0)` for?

packages/core/src/runtime/Settings.ts
phileba: why does it `undefined`? we can see if the `settings.duration = -1`, it's still is `-1` after the normalize function. if the value is a string, we will convert to the unit ms value.

packages/cli/src/cmd/run.ts
phileba: to stop the test right after finish all, sometimes it takes about ~10 seconds to stop the process and it is unnecessary.

packages/core/src/runtime/Settings.ts
HieuLeKMS: But what if the settings passed to `normalizeSettings` has `duration` = `undefined`?

packages/cli/src/cmd/run.ts
HieuLeKMS: You sure it's unnecessary? It might be something that's need to be done before exit, though. I think you should take another look

packages/core/src/runtime/Settings.ts
phileba: this case only happens when the user tries to input `duration = undefined` in the TestSetting, it does not make sense. But in case `duration = undefined`, the test will be run the same as `duration = -1`
```ts
if (settings.duration > 0) {
	this.timeout = setTimeout(() => {
		this.cancelled = true
	}, Number(settings.duration))
}
```
But I will add this check for this case.

packages/cli/src/cmd/run.ts
phileba: that is the end of the `handle` function means the element has done its job, maybe `yargs` need to do something inside. the behavior is:
- run without hook function: element stops immediately after the last step passed.
- run with hook function: element waits ~10 seconds to stop, need to investigate in another ticket.

HieuLeKMS: Then you should leave things as-is, and fix it later once you're done investigating

PR #362, opened 2020-11-25
Fix linting and formatting

andybluntish:
My editor was re-formatting every file on save. This fix ensures that ESLint formats code properly with Prettier.

- bump Prettier and ESLint plugins
- cleanup `.eslintignore` file to ensure we're skipping directories that don't need linting/formatting
- cleanup `.eslintrc.js` (remove redundancy)
- cleanup `.prettierrc` (remove rules that are the same as the defaults)
- add linting support for React (used in docs)
- add yarn scripts to run linting and formatting
- cleanup packages (de-dupe, add types)
- define prop-types for React components
- update eslint config so existing code passes
- update GH actions to be more consistent, and use LTS node
- run `eslint --fix` over the codebase


andybluntish:
Running with `--fix` changes a _lot_ of things, so going to wait until it's going to cause the least disruption before running that (e.g. after 2.0 is released).

andybluntish:
Replaced by https://github.com/flood-io/element/pull/366 which is based on the canary branch.

PR #368, opened 2020-12-02
feat: export and mock BaseLocator

nick-flood:
To enable users to create custom Locators


phileba (approved):

packages/element/index.ts
```diff
@@ -120,3 +120,5 @@ export {
  * @docPage helpers
  */
 export { RecoverWith } from '@flood/element-core'
+
+export { BaseLocator } from '@flood/element-core'
```
HieuLeKMS, line 124:
```suggestion
export { RecoverWith, BaseLocator } from '@flood/element-core'
```

PR #371, opened 2020-12-02
[Element 2.0] Change to logic to use the config file

honglax:
Ready to review.
Relate to the issue [FLO-2026](https://tricentisflood.atlassian.net/browse/FLO-2026)


packages/core/src/Runner.ts
```diff
@@ -128,7 +128,11 @@ export class Runner {
 						console.log(chalk.grey('--------------------------------------------'))
 					}
 					startTime = new Date()
-					console.log(`${chalk.bold('\u25CC')} ${iterationName} of ${this.looper.iterations}`)
+					console.log(
+						`${chalk.bold('\u25CC')} ${iterationName} of ${
+							this.looper.loopCount === -1 ? 'Infinity' : this.looper.loopCount
+						}`,
```
phileba, line 134: in case `loopCount = -1`, we should print out the console `Iternation <number>`, I confirmed with @sonnguyenmau

honglax: Got it

honglax: Resolved. Please check it again a @phileba.

packages/cli/src/cmd/run.ts
```diff
@@ -24,14 +24,20 @@ async function getAllTestScriptsFromConfiguration(
 ): Promise<RunCommonArguments> {
 	const fileErr = checkFile(args.configFile, 'Configuration file')
 	if (fileErr) throw fileErr
-	const { options, paths } = await readConfigFile(args.configFile)
+	const { options, paths, testSettings } = await readConfigFile(args.configFile)
```
HieuLeKMS, line 27: Currently, `readConfigFile()` returns `Promise<any>`. Could you make its return type more specific?

honglax: Resolved

packages/core/src/ElementOption.ts
```diff
@@ -35,6 +35,8 @@ export interface ElementRunArguments {
 	browser: BrowserType
 	export?: boolean
 	notExistingFiles: string[]
+	duration?: number
+	testSettings?: TestSettings
```
HieuLeKMS, line 39: Considering the configuration from config file is `ElementOptions`, should we not add these properties into `ElementRunArguments`?

honglax: We need to add `testSettings` as a **ElementRunArguments's** property for now. Gonna need an enhancement for this (separate settings from the config file and from CLI args) later.

HieuLeKMS (approved):

phileba:
actually, we change the constructor of the `Test.ts`, so please update in other class related likes `Test.spec.ts`, and it would be nice if we have a unit test for this.

packages/core/src/ElementOption.ts
```diff
@@ -57,6 +79,23 @@ export interface ElementOptions {
...
+export interface ElementConfig {
+	options: ElementOptions
+	paths: {
+		workRoot: string
+		testDataRoot: string
+		testPathMatch: string[]
+	}
+	flood?: {
+		hosted: boolean
+		vu: number
```
HieuLeKMS, line 91: It's actually `virtualUsers`, `vu` is just an alias for it in the CLI

honglax: I know it but @sonnguyenmau defined it in the template of Element Config. Therefore I use `vu` instead of `virtualUsers` to make sure they are consistent.

```
flood: {
    hosted: false,
    vu: 500,
    duration: 15,
    rampup: 0,
    regions: [''],
},
```

packages/cli/src/cmd/run.ts
```diff
@@ -19,19 +18,36 @@ import { resolve, dirname, basename, extname, join } from 'path'
...
-async function getAllTestScriptsFromConfiguration(
-	args: RunCommonArguments,
-): Promise<RunCommonArguments> {
-	const fileErr = checkFile(args.configFile, 'Configuration file')
+async function getConfigurationFromConfig(args: RunCommonArguments): Promise<RunCommonArguments> {
+	const { file, configFile } = args
+	const fileErr = checkFile(configFile, 'Configuration file')
 	if (fileErr) throw fileErr
-	const { options, paths } = await readConfigFile(args.configFile)
+	const configFileFromArgs: any = await readConfigFile(configFile)
```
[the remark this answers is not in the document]
honglax: Resolved

packages/core/src/ElementOption.ts
HieuLeKMS: Okay, then let's resolve this for now and test it later

packages/cli/src/cmd/run.ts
```diff
@@ -19,19 +19,42 @@ import { resolve, dirname, basename, extname, join } from 'path'
...
+		notExistingFiles,
+		runArgs: {
+			...args,
+		},
+		_,
+		$0,
+		file,
+		'fail-status-code': args['fail-status-code'],
+		configFile,
+		mu,
```
phileba, line 56: they are items in `args`, we already returned the `runArgs: { ... args }`, why we need to return them 1 more time?

honglax: Actually, the `args` may includes more properties (CLI flags). Because the return type of the function exclude the interface `Arguments`, I need to return these properties above.

phileba: yep, I guess that, so how about this one:
```ts
return {
    ...options,
    ...args,
    testSettings,
    testFiles,
    notExistingFiles,
   'fail-status-code': args['fail-status-code'],
}
```

phileba: I think we should avoid declaring much arg like `args`, `runArgs`..., it makes us confuses when reading the code :D

honglax: I used it for the first time of coding. But we have two types of running (single file and multiple files). Running a single test script allows using CLI flags but running multiple test scripts does not. Therefore, using like below
```
return {
    ...options,
    ...args,
}
```
does not fit for our purpose.

honglax: When using `return {
    ...options,
    ...args,
}`, the CLI flags always overwrite the ElementOptions in the config file (if they are defined). That is the reason I have to define another property named `runArgs`, check if run with single or multiple test scripts, and decide to use CLI flags or not.

honglax: Any suggestion for this issue?

phileba: it's okay.

phileba (approved):

PR #376, opened 2020-12-04
[Element 1.4] Cannot read property 'indexOf' of undefined" when running multiple test data

phileba:
resolve https://tricentisflood.atlassian.net/browse/FLO-2049

### DIFF
--- a/packages/core/src/test-data/Feeder.ts
+++ b/packages/core/src/test-data/Feeder.ts
@@ -52,9 +52,7 @@ export class Feeder<T> {
 		function checkWithTestKey(src: string[][], testKey: string[]): boolean {
 			return src.every((key: string[]) => {
 				if (key.length !== testKey.length) return false
-				return key.every((key: string) => {
-					return testKey.indexOf(key) >= 0
-				})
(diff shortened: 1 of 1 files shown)


packages/core/src/test-data/Feeder.ts
```diff
@@ -49,6 +49,15 @@ export class Feeder<T> {
 			filters.every(func => func(line, index, instanceID)),
 		)
 
+		function checkWithTestKey(src: string[][], testKey: string[]): boolean {
+			return src.every((key: string[]) => {
+				if (key.length !== testKey.length) return false
+				return key.every((key: string) => {
+					return testKey.indexOf(key) >= 0
+				})
```
honglax: Should we?
```
return key.every((key: string) => testKey.indexOf(key) >= 0)
```

honglax: I saw you've resolved it. Hence I'll resolve this conversation.

honglax (approved):

packages/core/src/test-data/Feeder.ts
phileba: you are too slow lol

PR #381, opened 2020-12-17
[Element 2.1] Show screenshot on terminal (iTerm)

phileba:
resolve https://tricentisflood.atlassian.net/browse/FLO-2027


packages/cli/src/cmd/run.ts
```diff
@@ -230,6 +230,10 @@ const cmd: CommandModule = {
 				describe: 'Export a HTML report after the test finished running',
 				type: 'boolean',
 			})
+			.option('--show-screenshot', {
+				describe: 'show screenshot in the terminal (iTerm only)',
+				type: 'boolean',
+			})
```
honglax, line 236: I think we should add one more property name `default` and set it `false` like
```
.option('--show-screenshot', {
    describe: 'show screenshot in the terminal (iTerm only)',
    type: 'boolean',
    default: false,
})
```

phileba: thank you for your suggestion, in this case, we only care if the `showScreenshot = true`, ignore the rest so the default value is `false` or `undefined` is the same.

honglax: Got it.

honglax:
Hi a @phileba, please resolve all conflicts before merging.

honglax (approved):

PR #384, opened 2020-12-22
Move typedoc to devDependency of element-core

rbayliss:
The `@flood/element-core` package currently has a vulnerable dependency (`highlight.js`) due to its dependency on the `typedoc` package.  Looking a little closer, it seems like `typedoc` is only used during development.  If I'm right about that, an easy solution is just to move it to `devDependencies`, and save everyone the extra packages.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -54,7 +54,6 @@
 		"source-map": "^0.7.0",
 		"strftime": "^0.10.0",
 		"term-img": "^2.1.0",
-		"typedoc": "^0.15.0",
 		"vm2": "^3.9.1",
 		"winston": "^3.0.0"
 	},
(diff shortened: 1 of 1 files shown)


phileba:
Hi @rbayliss , thank you for your update, please change the target branch to `flood-io:canary` then we can get it to merge.

rbayliss:
Ah, unfortunately, it looks like the canary branch is way ahead of `master`, which I started this change on.  To be clear though, while I appreciate that you want this in canary, our need is to get this into a 1.x release so we can resolve a security vulnerability without a major version update.  I'm willing to roll another PR for the canary branch, but it won't really help us unless it gets cherry-picked back to master and released as a 1.x tag.  Can you clarify what I should do here?

andy-flood:
Hi @rbayliss thanks for the PR, sorry it's taken a while to get to this.

Since you created this PR, the canary branch has been promoted to the main branch as version 2.x, so I have updated this PR to point to the 1.x branch as you mentioned you didn't want to upgrade just yet.

There is also an equivalent PR (https://github.com/flood-io/element/pull/472) to get this into 2.x.

If you're wanting to remain on the 1.x branch for a while longer, you can install the latest version using the `v1` npm tag:

```
npm install -g element-cli@v1
```

PR #388, opened 2020-12-26
[Element 2.0] Can not run a single test script without the config file

honglax:
This PR is going to fix the issue [FLO-2082](https://tricentisflood.atlassian.net/browse/FLO-2082).

### DIFF
--- a/packages/cli/src/cmd/run.ts
+++ b/packages/cli/src/cmd/run.ts
@@ -26,12 +26,10 @@ async function getConfigurationFromConfig(args: RunCommonArguments): Promise<Run
 	let notExistingFiles: string[]
 
 	if (fileErr) {
-		if (file) {
-			return { ...args, testFiles: [file], notExistingFiles: [] }
-		} else {
-			throw fileErr
-		}
+		if (!file) throw fileErr
+		return { ...args, testFiles: [file], notExistingFiles: [] }
 	}
+
 	const configFileFromArgs: ElementConfig = await readConfigFile(configFile)
 	const { options, paths, testSettings } = configFileFromArgs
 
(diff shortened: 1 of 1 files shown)


HieuLeKMS (approved):

packages/cli/src/cmd/run.ts
```diff
@@ -22,12 +22,19 @@ interface RunCommonArguments extends Arguments, ElementRunArguments {}
...
 	let testFiles: string[]
 	let notExistingFiles: string[]
 
+	if (fileErr) {
+		if (file) {
+			return { ...args, testFiles: [file], notExistingFiles: [] }
+		} else {
+			throw fileErr
+		}
+	}
```
phileba, line 34: In case run a test with a test script and no config file, why we don't keep this code line in the `handle` function and above this line `const result = await runSingleUser(runArgs)`
```ts
const runArgs = file ? args : await getConfigurationFromConfig(args)
```

packages/cli/src/cmd/run.ts
```diff
@@ -22,12 +22,19 @@ interface RunCommonArguments extends Arguments, ElementRunArguments {}
...
+	const configFileFromArgs: ElementConfig = await readConfigFile(configFile)
+	const { options, paths, testSettings } = configFileFromArgs
+
 	if (file) {
```
phileba, line 38: so in case run with the config file, we don't input file (always is undefined), so this condition never passed

packages/cli/src/cmd/run.ts
honglax: Because we need to get settings from the config file if possible.

packages/cli/src/cmd/run.ts
honglax: I can not understand your comment clearly. Can you explain more?

phileba: it's my fault, just confuse because I always thinking about the old logic, so now we combine running test with config file and test file into one function. I think we should apply the early return pattern like:
```ts
if (fileErr) {
	if (!file) {
		throw fileErr
	}
	return { ...args, testFiles: [file], notExistingFiles: [] }
}
```

packages/cli/src/cmd/run.ts
phileba: cool

packages/cli/src/cmd/run.ts
honglax: Yup, your code is more clearly than mine. Resolved.

phileba (approved):

Issue #380, opened 2020-12-16
Support the getCookie() API


tranphieu97:
**Is your feature request related to a problem? Please describe.**
When I working with the cookie in my Element project, I have to use the` evaluate()` function and the JS code to get the cookie string and some steps to the get the whole information of my cookie. 

**Describe the solution you'd like**
So, the idea is supporting an API for get the cookie properties by its name like 
`const myCookie = getCookie('cookieName')`

**Describe alternatives you've considered**

**Additional context**

honglax:
Hi @tranphieu97, new API `browser.getCookies()` has been released in Element@canary-17. Here is the documentation of this.
```
browser.getCookies(filterBy?: {urls?: string | string[], names?: string | string[]}): Promise<Cookie[]>

// Returned type
[
  {
      name: string,
      value: string,
      domain: string,
      path: string,
      expires: number, **Unix time in seconds**
      httpOnly: boolean,
      secure: boolean,
      sameSite: "Strict"|"Lax"|"None"
  }
]
```

Thank you for your suggestion.

PR #389, opened 2021-01-05
[Element 2.1] Implement new API browser.getCookies()

honglax:
Relate to this issue [FLO-2089](https://tricentisflood.atlassian.net/browse/FLO-2089) and this feature request [Issue #380](https://github.com/flood-io/element/issues/380)


HieuLeKMS (approved):

packages/core/src/page/types.ts
```diff
@@ -278,3 +278,8 @@ export interface ScreenshotOptions {
 	 */
 	encoding?: 'base64' | 'binary'
 }
+
+export interface CookiesFilterParams {
+	url?: string | string[] | undefined
+	name?: string | string[] | undefined
```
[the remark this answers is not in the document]
honglax: Got it

packages/core/src/runtime/Browser.ts
```diff
@@ -568,6 +570,21 @@ export class Browser<T> implements BrowserInterface {
...
+	public async getCookies(filterBy?: CookiesFilterParams): Promise<Cookie[]> {
+		const urls = filterBy?.urls
+		const names = filterBy?.names
+		let cookies = await this.page.context().cookies(urls)
+		if (names) {
+			cookies = [
+				...cookies.filter(cookie =>
+					typeof names === 'string' ? cookie.name === names : names.includes(cookie.name),
+				),
+			]
```
phileba, line 583:
```suggestion
	cookies = cookies.filter(cookie =>
		typeof names === 'string' ? cookie.name === names : names.includes(cookie.name),
	),
```

honglax: I don't think so. Because the param `names` can be `string` or `string[]` and we need to make sure that the filtered cookies are correct with the `filterByName` param.

For example, the browser has a cookie named `largest_order` and its value is `11`. This LOC below shows us how to get the correct cookie named `largest_order`
```
await browser.getCookies({names: 'largest_order' })

// or
await browser.getCookies({names: ['largest_order'] })
```

But if we change the filter source code to `cookies = cookies.filter(cookie => names.includes(cookie.name))`, when the end-users put the names param is `largest_order-dumb-thing`, they also get cookie named `largest_order`.

phileba: okay, so just keep the old check, and do we need the spread syntax?

honglax: Oh, that's my old habits. Gonna remove it in the next commit.

phileba (approved):

PR #395, opened 2021-01-19
[Element 1.4] Release Element 1.4

honglax:


nick-flood:
Looks good, let's hold off on merge until we get a 👍  from QA. Should be by end of week

PR #397, opened 2021-01-20
[Element 2.1] Enhancement terminal report

phileba:
resolve https://tricentisflood.atlassian.net/browse/FLO-2040


packages/core/src/interface/ITest.ts
```diff
@@ -35,7 +35,6 @@ export interface ITest {
 		testDataRecord: any,
 	): Promise<void>
 	doStepDelay(): Promise<void>
-	willRunCommand(testObserver: TestObserver, browser: Browser<Step>, command: string): Promise<void>
```
honglax, line 38: Why did you remove `willRunCommand()`?

IMO, it should be changed to 
```
willRunCommand(
    testObserver: TestObserver,
    browser: Browser<Step>,
    command: string,
    args: string,
): Promise<void>
```

packages/core/src/runtime/Test.ts
```diff
@@ -255,6 +257,12 @@ export default class Test implements ITest {
 				}
 			})
 			await this.afterRunSteps(stepIterator)
+			if (this.settings.showScreenshot && process.env.TERM_PROGRAM === 'iTerm.app') {
+				for (const path of browser.fetchScreenshots()) {
+					console.log(basename(path))
```
honglax, line 262: Do we need to keep this `console.log()`

packages/core/src/interface/ITest.ts
phileba: okay, it's my fault lol

packages/core/src/runtime/Test.ts
phileba: yes, we need to show the file name of the screenshot

honglax (approved):

PR #409, opened 2021-01-27
fix: unknown option --disable-gpu

HieuLeKMS:
Fix error: `Cannot parse arguments: Unknown option --disable-gpu` when running a test with `webkit` browser, on a flood-element docker container

### DIFF
--- a/packages/core/src/driver/Playwright.ts
+++ b/packages/core/src/driver/Playwright.ts
@@ -116,6 +116,10 @@ export async function launch(
 		...passedOptions,
 	}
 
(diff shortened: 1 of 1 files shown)


packages/core/src/driver/Playwright.ts
```diff
@@ -116,7 +116,6 @@ export async function launch(
 		...passedOptions,
 	}
 
-	options.args.push('--disable-gpu')
```
phileba, line 119: I think we just remove this arg if the browser is webkit?

phileba (approved):

PR #424, opened 2021-02-05
[Element 2.0] Support Chrome stable when running on Flood

phileba:


packages/core/src/driver/Playwright.ts
```diff
@@ -127,9 +135,21 @@ export async function launch(
...
+		browserType = 'chromium'
+		switch (process.platform) {
+			case 'darwin':
+				options.executablePath = '/Applications/Google Chrome.app/Contents/MacOS/Google Chrome'
+				break
+			case 'win32':
+				options.executablePath = 'C:/Program Files (x86)/GoogleChrome/Applicationchrome.exe'
+				break
+			default:
+				options.executablePath = '/usr/bin/google-chrome-stable'
```
honglax, line 150: What happens if the end-user install Google Chrome at their own path (not default path)?

phileba: they need to use the `executablePath`:
```
browserLaunchOptions: {
	executablePath: 'custom_path',
},
```

phileba: the purpose of this part is to support end-user who wants to run the test with the chrome stable on Flood

honglax: The executablePath of Chrome in Flood Grid is `'/usr/bin/google-chrome-stable'`? Am I correct?

HieuLeKMS (approved):

honglax (approved):
LGTM.

PR #436, opened 2021-02-19
fix: correct link in test data guide

leeomara:
The "Flood challenge with test data" in the "Using Test Data" guide had the wrong syntax and didn't point anywhere.

This was in the "More information" section at the very bottom.
https://element.flood.io/docs/guides/test-data#more-information

This change point "Flood challenge with test data" to the example script at https://github.com/flood-io/element/blob/master/examples/flood-challenge.ts


andy-flood:
Thank you for the fix, sorry for the delay!

PR #437, opened 2021-02-22
[Element 2.0] Fix test that can not show image on iTerm2 when folder has the element config file

honglax:
Resolve issue [FLO-2172](https://tricentisflood.atlassian.net/browse/FLO-2172).

### DIFF
--- a/packages/core/src/ElementOption.ts
+++ b/packages/core/src/ElementOption.ts
@@ -204,7 +204,7 @@ export function normalizeElementOptions(
 
 	const reporter = verboseBool ? new VerboseReporter(spinnies) : new BaseReporter(spinnies)
 
-	const opts: ElementOptions = {
+	let opts: ElementOptions = {
 		testScript: file,
 		strictCompilation: args.strict ?? false,
 		reporter: reporter,
@@ -221,7 +221,7 @@ export function normalizeElementOptions(
 		export: args.export,
 		executablePath: args.executablePath ?? '',
 		downloadsPath: args.downloadsPath ?? '',
-		showScreenshot: args.showScreenshot,
+		showScreenshot: args.showScreenshot ?? false,
 	}
 
 	opts.testSettings = { ...testSettings }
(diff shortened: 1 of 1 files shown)


packages/core/src/ElementOption.ts
```diff
@@ -246,13 +246,7 @@ export function normalizeElementOptions(
...
 		if (loopCount !== undefined) opts.testSettingOverrides.loopCount = loopCount
 		if (duration !== undefined) opts.testSettingOverrides.duration = duration
-		opts.headless = runArgs.headless ?? opts.headless
-		opts.devtools = runArgs.devtools ?? opts.devtools
-		opts.sandbox = runArgs.sandbox ?? opts.sandbox
-		opts.export = runArgs.export ?? opts.export
-		opts.browser = runArgs.browser ?? opts.browser
-		opts.executablePath = runArgs.executablePath ?? opts.executablePath
-		opts.downloadsPath = runArgs.downloadsPath ?? opts.downloadsPath
+		opts = { ...opts, ...runArgs }
```
HieuLeKMS, line 249: With this change, if `runArgs` doesn't have one of these properties, for example `headless`, then `opts.headless` won't be changed, but if `runArgs` does have the property `headless`, but its value is `undefined`, then `opts.headless` will be set to `undefined`, whereas with the old code, `opts.headless` won't be changed in both case.

honglax: If the end-user does not provide any arguments, the runArgs does not have that property. And this scope is executed when there is config file in the folder. 

If you and a @phileba wanna change it back to the old one and add a new line to check showScreenshot, I’m happy to do it.

nick-flood (approved):

packages/core/src/ElementOption.ts
HieuLeKMS: If you are aware and sure there're no problems with it, then no need to change back

HieuLeKMS (approved):

phileba (approved):

packages/core/src/ElementOption.ts
honglax: Got it.

PR #459, opened 2021-03-24
[Element 1.x.x] Resolve unhandled rejection on Windows and Ubuntu

honglax:
Resolve this issue [**FLO-2208**](https://tricentisflood.atlassian.net/browse/FLO-2208)

Just a little fixing base on the usage of `command-exists` [here](https://www.npmjs.com/package/command-exists).

### DIFF
--- a/packages/cli/src/utils/updateCheck.ts
+++ b/packages/cli/src/utils/updateCheck.ts
@@ -13,7 +13,14 @@ import { prerelease, major, minor, patch } from 'semver'
 type Update = { latest: string }
 
 function brewUpdateMessage(version: string, distTag: string, update: Update): string | undefined {
-	const brew = commandExists('brew')
+	let brew = false
+	commandExists('brew', function(err, isBrewExists) {
+		if (err) {
+			console.error(err)
+		} else {
+			brew = !!isBrewExists
+		}
+	})
 	if (__dirname.includes('Cellar') && brew) {
 		let brewSpec: string
 		if (distTag === 'latest') {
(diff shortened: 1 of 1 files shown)


packages/cli/src/utils/updateCheck.ts
```diff
@@ -33,9 +40,18 @@ function yarnUpdateMessage(
...
+		if (err) {
+			console.error(err)
+			result = undefined
+		} else {
+			result = isYarnExists
+				? chalk`Get it by running {greenBright yarn global upgrade element-cli@${distTag}}`
+				: undefined
+		}
+	})
+	return result
```
nick-flood, line 54: is commandExists asynchronous? Looks like it from the callback, won't we get order of execution errors?

nick-flood (changes requested):

nick-flood:
Rather than using the async version and having to re-write the function signature i think we should use the sync version `commandExists.sync`

honglax:
> Rather than using the async version and having to re-write the function signature i think we should use the sync version `commandExists.sync`

I'll give it a try.

honglax:
> Rather than using the async version and having to re-write the function signature i think we should use the sync version `commandExists.sync`

Hi @nick-flood, I've changed to the sync version and it works well. Please re-review this PR. Thank you.

packages/cli/src/utils/updateCheck.ts
honglax: Hi @nick-flood, I've changed to the sync version and it works well. Please re-review this PR. Thank you.

phileba (approved):

HieuLeKMS (approved):

nick-flood (approved):

PR #460, opened 2021-03-25
fix: fail to start 2nd iteration when running on Flood

HieuLeKMS:
Apply custom console when running on Flood, which contains the missing method `setGroupDept`

### DIFF
--- a/packages/core/src/Element.ts
+++ b/packages/core/src/Element.ts
@@ -13,7 +13,10 @@ import { ElementResult } from './ElementResult'
 import { join } from 'path'
 import findRoot from 'find-root'
 
-export async function runSingleTestScript(opts: ElementOptions): Promise<IterationResult[]> {
+export async function runSingleTestScript(
+	opts: ElementOptions,
+	spinnies?: any,
+): Promise<IterationResult[]> {
 	const {
 		testScript,
 		clientFactory,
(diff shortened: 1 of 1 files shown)


honglax (approved):
LGTM 🚢

packages/core/src/Element.ts
```diff
@@ -27,6 +30,12 @@ export async function runSingleTestScript(opts: ElementOptions): Promise<Iterati
 		showScreenshot,
 	} = opts
 
+	global.console = new CustomConsole(
+		process.stdout,
+		process.stderr,
+		spinnies ? spinnies : new Spinnies(),
```
phileba, line 36:
```suggestion
		spinnies || new Spinnies(),
```

phileba (approved):

packages/core/src/Element.ts
```diff
@@ -27,6 +30,8 @@ export async function runSingleTestScript(opts: ElementOptions): Promise<Iterati
 		showScreenshot,
 	} = opts
 
+	global.console = new CustomConsole(process.stdout, process.stderr, spinnies || new Spinnies())
```
nick-flood, line 33: Feels kinda gross modifying the global console, is this necessary?

HieuLeKMS: That was to add the method `setGroupDept` to the console. I think it's safe since the custom console extends the global one.

HieuLeKMS: I'm not sure if it's necessary, but I think it's better to manage just one logger.

nick-flood (approved):

PR #570, opened 2022-03-02
fix: guard against negative response time measurements

kunzmann:
Backporting https://github.com/flood-io/element/pull/552 to element 1.5


lachie-flood (approved):
:shipit:

PR #575, opened 2022-03-28
Update faker to use new 'official' version

fr0th:
see: https://www.npmjs.com/package/@faker-js/faker


alexthomsonnz:
+1 This, would be great to see this merged soon.

Issue #624, opened 2023-07-21
Send the User Worker Identifier/Name along with the step parameter


arvind-fluxon:
**Is your feature request related to a problem? Please describe.**
Yes, many times when we run "stage" based execution with a `loopCount` of `1`, we end up not getting the data rolled over, since it depends on the iteration.

In order for us to have max parallelism, sometimes in the stage we try with a target of 100 users. 
Now, each of these workers are unique as of Element, however, the actual code within behaves and gets the same data (from TestData) and even if we expose a different generator function, they cannot get unique values in each run.
As an alternate, its better to know the worker identifier within the step to pick and choose what's needed.

PR #625, opened 2023-07-21
feat: [ISSUE-624] Send worker name as part of step execution.

arvindkumarc:
This helps to have more control over the workers and take better data slicing even with single iteration and parallel running loads.

Consider the following example: 
```
  loopCount: 1,

  stages: [
    {
      duration: `2m`,
      target: 5,
    },
  ],
};
```

We expect 5 parallel workers being launched, and there is no unique way to identify these workers.
The `TestData` `circular()` or `shuffle` will not help here since they work based on the `loopCount`. `shuffle` to an extent but with duplicates.

With this approach, we will be able to slice the data based on the launched workers and control which to be picked.

More details in #624 

Thank you!

### DIFF
--- a/packages/core/src/runtime/Test.ts
+++ b/packages/core/src/runtime/Test.ts
@@ -395,7 +395,7 @@ export default class Test implements ITest {
 		try {
 			debug(`Run step: ${step.name}`) // ${step.fn.toString()}`)
 			browser.settings = { ...this.settings, ...step.options }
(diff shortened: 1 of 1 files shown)


vipulsharma144:
@arvindkumarc Thanks for the fix However i tried installing via your fork and it seems to give me undefined with `--mu` . Can you provide some steps to follow please . 
Thanks

